PR #91, opened 2021-11-15
do not put form element within modal layout

### LINKED ISSUE #90: edit file dialog is not scrollable.
Reproduce:

* open publication
* go to tab "full text"
* click on the ".." on right of a file, and choose option "Edit.."
* a modal appears, allowing you to edit and save file metadata

Problems:

* if your screen is not height enough, you cannot see or access the top and bottom (where the save and cancel button is!), nor can you scroll to it.

Screenshot 2021-11-15 at 12 20 22

nicolasfranck:
Fixes #90


nics (changes requested):
putting the form outside the modal is probably not a good idea, this also fixes the issue:
```
<form class="modal-content">
  <div class="modal-header">
  ...
</form>
```

nicolasfranck:
updated

nics (approved):

PR #94, opened 2021-11-15
Fix issue 90: use hx-include instead of wrapping form element.

netsensei:
I'm using `hx-include` instead of wrapping `modal-body` and `modal-footer` in a wrapping `form` element to achieve the exact same behavior.

See: https://htmx.org/attributes/hx-include/

Advantage: no weird HTML constructs just to keep catering to HTMX's default behavior of including any nearby form elements into a POST request. `hx-include` was purposefully designed as an escape hatch for this exact use case.


nicolasfranck:
@netsensei the missing piece here is not another regular input value, but the submit button that causes the form to be submitted..

nics (approved):
more idiomatic

PR #227, opened 2021-12-10
Fix issue 209

### LINKED ISSUE #209: Publish to biblio returns json  error
If you try to publish a publication/dataset that has not yet got a `title`,
then you get a json error.

Reason:
  * attribute `title` is only required when `status` is `public` (validation). So users are not aware of this "problem" until the last moment.
  * button `Publish to biblio` is only a link that updates the publication,
     but it has no way to express form errors when they should appear. 
     This is problematic even more when dealing with a button like `Publish all to biblio`
     in the wizard, when one or more publications lack a title.

nicolasfranck:
fixes #209


internal/controllers/datasets.go
```diff
@@ -99,11 +99,32 @@ func (c *Datasets) Show(w http.ResponseWriter, r *http.Request) {
...
+	oldStatus := dataset.Status
 	savedDataset, err := c.Engine.PublishDataset(dataset)
 	if err != nil {
-		log.Println(err)
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+
+		// pub not updated, but status has been set in struct
+		savedDataset = dataset
+		savedDataset.Status = oldStatus
```
nics, line 110: this should by handled by PublishDataset, not the controller

internal/controllers/datasets.go
```diff
@@ -99,11 +99,32 @@ func (c *Datasets) Show(w http.ResponseWriter, r *http.Request) {
...
+
+		if e, ok := err.(jsonapi.Errors); ok {
+
+			flashes = append(flashes, views.Flash{Type: "error", Message: e[0].Title})
+
+		} else {
+
+			log.Println(err)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
nics, line 119: this should also be a 'could not publish' flash message

nics (approved):

PR #264, opened 2022-01-24
add show builder link; link to doi publisher website

### LINKED ISSUE #195: link out to publication at publisher's website
**Describe the improvement**
Not sure whether this is already implemented and currently not working.
As for datasets, I want to be able to view the publication on the publisher's website by linking out using the DOI (if available).

I want to view the publisher's website in a new tab.

nicolasfranck:
Fixes #195

### DIFF
--- a/internal/views/show_builder.go
+++ b/internal/views/show_builder.go
@@ -119,6 +119,10 @@ func (b *ShowBuilder) Tooltip(v string) showOption {
 	}
 }
 
+func (b *ShowBuilder) Link(opts ...showOption) (template.HTML, error) {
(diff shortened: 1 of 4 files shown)


nics (commented):
I would turn this into a showOption for the existing Text builder, e.g.

`(.FormatLink "https://dx.doi.org/%s")`

And interpolate the value

nicolasfranck:
@nics in that case I would either need a "if-else" inside the template, or one in that FormatLink, as the doi can be empty. That is why I choose to use a template, which gives the freedom to add for example an icon.

nics:
replaced by 4bb352a5203925dc4dde8b310432223153ce4476

PR #371, opened 2022-05-24
add facet faculty

nicolasfranck:
I've created a separate field `faculty` where all values from `department.tree._id` are copied to.
That makes it easier to go from `f[faculty]=LW` to a filter on `department.tree._id` (which
would be another exception to be added to the parseScope code).

Need to reindex.

### DIFF
--- a/etc/es6/publication.json
+++ b/etc/es6/publication.json
@@ -173,6 +173,7 @@
                 "date_updated": {
                     "type": "date"
                 },
+                "faculty": { "type": "keyword" },
                 "department": {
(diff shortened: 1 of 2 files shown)


nicolasfranck:
Or we could just add `faculty` as a field, to be created automatically in that new pipeline. That would simplify the filters and facets..

nics:
thanks @nicolasfranck
agree that this is dirty, we can clean it up later
for example by adding a type to the organization info `{id: 123, type: "faculty"}`?

PR #423, opened 2022-06-25
Handlers dataset publish

netsensei:
Includes:

* HX-Redirect redirection on dataset publish.
* Updated validation rules for dataset model.
* Session based Flash messages.


nics (commented):
gorilla sessions already has flash helpers: https://github.com/gorilla/sessions/blob/master/sessions.go#L43
can't we use those?

netsensei:
Done.

nics (approved):

PR #427, opened 2022-06-27
Handlers - publicationviewing

nicolasfranck:
* complete description tab for publications
* added routers for description, files, contributors and datasets. Except for "description" those tabs return almost empty page. Just to make golang happy.
* changed router names that display modal with departments and projects, well also to make golang happy, as I needed to add a link on the "edit" button
* reorganized description tab:
  * card do not need "id"
  * card-body has been given an "id", always "<section->body". e.g. "publication-details-body"
  * "id" is not given directly to div with class "card-body", but actually to a wrapping div. This way a template like "abstract_body.gohtml" can choose between a card-body with zero or more padding depending on its content ("no abstracts"). The wrapping div with the id is the target of htmx after update from a modal.

What could be better:
* use of `Locale` when executing the `ValueTemplate` in a Display object. Now I had to manually prepare the language names


internal/app/displays/display_conference.go
```diff
@@ -0,0 +1,31 @@
+package displays
+
+import (
+	"github.com/ugent-library/biblio-backend/internal/locale"
+	"github.com/ugent-library/biblio-backend/internal/models"
+	"github.com/ugent-library/biblio-backend/internal/render/display"
+	"github.com/ugent-library/biblio-backend/internal/services/webapp/helpers"
+)
+
+func DisplayConference(l *locale.Locale, p *models.Publication) *display.Display {
```
nics, line 10: I would only pass the conference, not the whole publication

internal/app/displays/display_type_conference.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"github.com/ugent-library/biblio-backend/internal/locale"
+	"github.com/ugent-library/biblio-backend/internal/models"
+	"github.com/ugent-library/biblio-backend/internal/render/display"
+	"github.com/ugent-library/biblio-backend/internal/services/webapp/helpers"
+)
+
+func DisplayTypeConference(l *locale.Locale, p *models.Publication) *display.Display {
+	//TODO: better way to do this?
```
nics, line 11: I'll make a helper

internal/app/handlers/publicationviewing/show.go
```diff
@@ -0,0 +1,162 @@
...
+
+var allowedSubNavs = []string{
+	"description",
+	"files",
+	"contributors",
+	"datasets",
+}
+
+func (h *Handler) Show(w http.ResponseWriter, r *http.Request, ctx Context) {
+	searchArgs := models.NewSearchArgs()
```
nics, line 52: i'll move this to the context

internal/app/handlers/publicationviewing/show.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := bind.Request(r, searchArgs); err != nil {
+		render.BadRequest(w, r, err)
+		return
+	}
+
+	render.Render(w, "publication/show_description", YieldShowDescription{
+		Context:               ctx,
+		ActiveSubNav:          "description",
+		SearchArgs:            searchArgs,
+		DisplayDetails:        displayDetails(ctx),
```
nics, line 84: i'll move the switch to the displays package

nicolasfranck: switch?

internal/services/webapp/routes/register.go
```diff
@@ -378,9 +383,26 @@ func Register(services *backends.Services, oldBase controllers.Base, oidcClient
 	pubPublishRouter.Use(middleware.RequireCanPublishPublication)
 	pubDeleteRouter := pubRouter.PathPrefix("").Subrouter()
 	pubDeleteRouter.Use(middleware.RequireCanDeletePublication)
-	pubRouter.HandleFunc("", publicationsController.Show).
+	r.HandleFunc("/publication/{id}",
```
nics, line 386: todo: move routes to the top with the others

internal/app/handlers/publicationviewing/show.go
nics: displayDetails()

PR #568, opened 2022-09-06
fix issue 529

### LINKED ISSUE #529: I can filter on the publishing year of a record (publications and datasets)
# Story

I can filter on the publishing year of publication or dataset so I can see which records need follow-up in time.

Filters depend on the publication years that have been registered

# Context old Biblio
Image
Image

## Success criteria
- [x] Available for publications
- [x] Available for datasets
View other success criteria on https://github.com/ugent-library/biblio-backend/issues/519

# Implementation and context
View design and components on https://github.com/ugent-library/biblio-backend/issues/519

nicolasfranck:
Fixes #529 

Rather a poor man's solution:

* shows a static list of available years in the index, in sorted order. List may grow unruly. 
* what if there are more than 200 unique values for year available in the index? That is the maximum number of facet values fetched from the index

### DIFF
--- a/client/auth/basic.go
+++ b/client/auth/basic.go
@@ -1,0 +1,23 @@
+package auth
+
+import (
+	"context"
+	"encoding/base64"
+)
+
+type BasicAuth struct {
+	User     string
+	Password string
+}
+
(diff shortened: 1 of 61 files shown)


nics:
@nicolasfranck i think you can return up to 999 values
indeed not ideal, but we can design a better widget later

PR #587, opened 2022-09-09
Fix issue 524

### LINKED ISSUE #524: I can filter on records that are "with message" or "without message" (publications and datasets)
# Story
I can filter on records that have a message or do not have a message, so I can see which records need follow-up.

# Context old Biblio
Image

## Success criteria
- [ ] Available for publications
- [ ] Available for datasets
View other success criteria on https://github.com/ugent-library/biblio-backend/issues/519

# Implementation and context
View design and components on https://github.com/ugent-library/biblio-backend/issues/519

nicolasfranck:
* fixes #524 
* added model `IndexedPublication` and `IndexedDataset` to make derived fields possible.
* fixed templates crashes when rendering facets for  `/dataset` due to zero slices in `hits.Facets`. The culprit is that `hits.Facets[facetName]` can be a zero slice, when the associated buckets in ES did not have values.


nics (commented):
@nicolasfranck I was thinking about a generalized solution with a `fields` field like the biblio frontend

```
fields:["message"]
```

what do you think?

nics (approved):

PR #591, opened 2022-09-09
format all date as UTC, with millisecond precision

### LINKED ISSUE #588: Make time formatting during serializing and deserializing uniform
**Describe the bug**

During import of JSON from librecat
the `date_created` is formatted in Zulu time,
without millisecond or nanosecond precision,
while `date_updated` does have that precision.

Moreover, during indexing the millisecond precision
is lost during formatting, due to formatting as RFC3339. So if you have a lot of documents
that originate from a fast import (and so are created
within the same second), the sorting will depend..

Literally spoken, this is no technical issue,
but it leads to two types of time formatting

**Expected behavior**

Use this format of serializing:

* UTC Zulu time. It makes sorting for other applications simpler (no interpretation needed)
* Include millisecond precision

nicolasfranck:
fixes #588

### DIFF
--- a/internal/backends/es6/indexed_dataset.go
+++ b/internal/backends/es6/indexed_dataset.go
@@ -1,25 +1,31 @@
 package es6
 
 import (
-	"time"
-
 	"github.com/ugent-library/biblio-backend/internal/models"
 )
 
 type IndexedDataset struct {
 	models.Dataset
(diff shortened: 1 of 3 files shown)


nics:
@nicolasfranck weird that there are helper methods for second and nano, but not for milli

PR #592, opened 2022-09-09
fix issue 466

### LINKED ISSUE #466: Data format dropdown list free text option does not work as expected
# Description bug

When trying to add non-predefined a data format:
- the dropdown still tries to match one of the items in the list, if it found a similar match
- the dropdown does not go away
- it does not disappear after adding a new data format

## Steps to reproduce

1. Go to eliot https://eliot1.ugent.be/biblio-backend/dataset/
2. Click "Edit" on the "Dataset details" inside "Description"
3. Add values that match partially with suggested data formats (e.g. key or new)

It does not happen when you try to add a keyword that does not match at all:
Image

### Example 1: Adding a data format "key"
One cannot enter "key"
Image

### Example 2:  Clicking the plus sign after adding "key"
Image

### Example 3: Adding a data format "new" after adding "key"
Image

### Example 4: Adding another random data format after adding the previous examples
Image

### Example 5: matching one of the predefined "new" options
Image

# Expected behaviour
- You can add the "new" value entered
- The dropdown goes away when you add a new keyword

## Suggestion
Add "new" value to make action explicit at the bottom of the list

Image

### Code example
`<li class="list-group-item list-group-item-action d-flex justify-content-between align-items-center autocomplete-hit" data-value="application/pgp-keys"><strong>Add new data format "Key"</strong></li>`

# System information

## Desktop
 - OS: macOS Monterey
 - Browser: Safari (Version 15.5), Chrome (latest), Firefox (latest)

nicolasfranck:
fixes #466 

@mietcls I've used the text `Add custom data format` instead of `Add new data format` as that last suggests that you're updating an existing vocabulary.

### DIFF
--- a/internal/app/handlers/mediatypes/handler.go
+++ b/internal/app/handlers/mediatypes/handler.go
@@ -34,7 +34,8 @@ func (h *Handler) Wrap(fn func(http.ResponseWriter, *http.Request, Context)) htt
 
 type YieldSuggest struct {
 	Context
-	Hits []models.Completion
+	Hits  []models.Completion
+	Query string
 }
 
 func (h *Handler) Suggest(w http.ResponseWriter, r *http.Request, ctx Context) {
@@ -55,5 +56,6 @@ func (h *Handler) Suggest(w http.ResponseWriter, r *http.Request, ctx Context) {
 	render.View(w, "media_types/suggest", YieldSuggest{
 		Context: ctx,
 		Hits:    hits,
+		Query:   q,
 	})
 }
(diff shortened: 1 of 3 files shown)


mietcls:
@nicolasfranck for my understanding, if people add a custom one it is only visible to themselves, correct? So I would not be able to choose the new data format you added?

nicolasfranck:
@mietcls it means that the key they entered isn't drawn from an existing vocabulary. The format is by all means a string, that is mostly drawn from an existing list, and by exception not. So no "own list". That is why the word "add" feels a bit off to me. Why not `Use custom data format "my custom format"`

mietcls:
@nicolasfranck My question comes down to this: is this custom value added to an existing vocabulary afterwards? Can others use that custom value?

If I understand correctly, the answer is no. `Use custom data format "my custom format"` is a good suggestion in that case.

nics (approved):

Issue #531, opened 2022-09-01
Filter on created since


mietcls:
# Story

I can filter on records created since a specific time.

# Implementation

## Success criteria

- [ ] I can filter on records "created since", by entering:
  - [ ] today
  - [ ] yesterday
  - [ ] YYYY
  - [ ] YYYY-MM
  - [ ] YYYY-MM-DD
  - [ ] or iso utc timestamp YYYY-MM-DDTHH:MM:SST
- [ ] If the entered date does not work, I get an error state https://github.com/ugent-library/biblio-backend/issues/544

### Out of scope
- Making the input case-insensitive
- Reversing the date notation
- Other types of dates

## Design
<img width="325" alt="Schermafbeelding 2022-09-02 om 12 59 26" src="https://user-images.githubusercontent.com/1572885/188126271-21e100d7-4764-4893-8d14-f7561da53ce6.png">

## Example code

_Warning: edited out the positioning and placement._

```
<div class="dropdown-menu show" x-placement="bottom-start">
  <div class="bc-navbar bc-navbar--bordered-bottom">
    <div class="bc-toolbar bc-toolbar--auto">
      <div class="bc-toolbar-left">
        <h4 class="text-nowrap">Created since</h4>
      </div>
    </div>
  </div>
  <div class="p-6">
    <label class="col-form-label">Show records created since</label>
    <input class="form-control" type="text" placeholder="">
    <small class="form-text text-muted">Type a date (DD/MM/YYYY), year (YYYY) or timing (yesterday).</small>
  </div>
  <div class="bc-navbar bc-navbar--large">
    <button class="btn btn-primary btn-block" type="button">Apply filter</button>
  </div>
</div>
```

# Context

## Old Biblio
![Schermafbeelding 2022-09-01 om 18 20 46](https://user-images.githubusercontent.com/1572885/187963965-f8bd4a93-23a0-4aff-b3d7-787aef9400d0.png)

mietcls:
@nics what are the rules you use?

nics:
@mietcls current rules are:

- 'today'
- 'yesterday'
- YYYY
- YYYY-MM
- YYYY-MM-DD
- or iso utc timestamp YYYY-MM-DDTHH:MM:SST

nics:
done

PR #602, opened 2022-09-16
filter on dates since for date_created and date_updated

### LINKED ISSUE #530: Filter on updated since

nicolasfranck:
Fixes #530 and #531

Notes:

* I added a method `SearchArgs#Cleanup` that removes empty slices, or empty values from the filters. Reason: this type of filter introduces empty values in the url as it is a simple text input, while the other filters work with checkboxes that do not send anything when unchecked. The cleanup is called in both handlers of publications and datasets.
* This is a "new" kind of filter, for which no facets are needed
* Since where are filtering here not just on regular values, but on dates, query syntax errors may occur. As I did not see a way to report this errors, I "fixed" it for now by searching for a date in the far future so it always returns zero results. See `es6.ParseTimeSince`. Maybe not ideal, but expected.

### DIFF
--- a/internal/app/handlers/datasetsearching/handler.go
+++ b/internal/app/handlers/datasetsearching/handler.go
@@ -33,6 +33,7 @@ func (h *Handler) Wrap(fn func(http.ResponseWriter, *http.Request, Context)) htt
 			render.BadRequest(w, r, err)
 			return
 		}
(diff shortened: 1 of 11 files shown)


nicolasfranck:
TODO: return query syntax error instead of searching for something in the future.

PR #707, opened 2022-10-13
Publication create handle

### LINKED ISSUE #706: hook to generate handle id for public records

nicolasfranck:
Fixes #706

* Create handle via handle-server-api on (re)publish, and set the attribute `handle` in publication
* Removes handle via handle-server-api on withdraw, and removes the attribute `handle` from publication. Not sure if that last is necessary

To make this work, add the following variables:

```
# base url of handle server
export BIBLIO_BACKEND_HDL_SRV_URL="http://localhost:3000/handles"
export BIBLIO_BACKEND_HDL_SRV_PREFIX="1854"
export BIBLIO_BACKEND_HDL_SRV_USERNAME="handle"
export BIBLIO_BACKEND_HDL_SRV_PASSWORD="handle"
```

Notes:

* the handle server client built here assumes that the internal handle should be built as follows: `<base-url>/<prefix>/LU-<id>`. e.g. http://localhost:3000/handles/1854/LU-1234
* the handle server client built here assumes that the external handle should be built as follows: `http://hdl.handle.net/<prefix>/LU-<id>`. e.g. http://hdl.handle.net/1854/LU-1234
* the pipeline `publication.PublishPipeline` and `publication.UnpublishPipeline` probably deserve a better place to call. Now they are only called from within the http controllers.


nicolasfranck:
Update: now you can disable updates to the handle server:

```
export BIBLIO_BACKEND_HDL_SRV_ENABLED=true
```

disabled by default if this variable isn't set

TODO:

* failures are now reported in the log, but not acted upon. Stop someone from saving?

nics (approved):

PR #759, opened 2022-10-26
add router that clears embargo for publications and datasets

nicolasfranck:
add routers to clear embargo for publications and datasets


nics (commented):
routes need to be secured with BasicAuth

nicolasfranck:
fixed

PR #791, opened 2022-10-31
Extra fixes

nicolasfranck:
* add attribute `HasBeenPublic` to models Publication and Dataset, and make sure the attribute is set automatically
* add missing fixes in handler frontoffice
* return error `ErrNotFound` from repository when record is not found, so handlers can distinguish between "internal server error" and "not found". For some cases, like associating a publication to a dataset, I left it there as that makes sense to issue an internal server error (should actually be a 400, but ok).


internal/app/handlers/publicationediting/handler.go
```diff
@@ -39,13 +39,23 @@ func (h *Handler) Wrap(fn func(http.ResponseWriter, *http.Request, Context)) htt
...
+					"user", ctx.User.ID,
+				)
+				render.InternalServerError(w, r, err)
+			}
 			return
 		}
 
 		if !ctx.User.CanEditPublication(pub) {
-			h.Logger.Warn("edit dataset: user isn't allowed to edit the dataset:", "errors", err, "publication", id, "user", ctx.User.ID)
+			h.Logger.Warn("edit publication: user isn't allowed to publication the dataset:", "errors", err, "publication", id, "user", ctx.User.ID)
```
nics, line 58: typo

internal/app/handlers/frontoffice/handler.go
```diff
@@ -35,7 +36,7 @@ var licenseMap = map[string]string{
 	"CC-BY-NC-ND-4.0":  "Creative Commons Attribution-NonCommercial-NoDerivatives 4.0 International Public License (CC BY-NC-ND 4.0)",
 	"InCopyright":      "No license (in copyright)",
 	"LicenseNotListed": "The license is not listed here",
-	"CopyrightUnknown": "I don't know the status of the copyright of this publication",
+	"CopyrightUnknown": "I don't know the status of the copyright for this publication",
```
nics, line 39: origineel is correct

nicolasfranck: I wasn't sure, but there was a difference between the frontoffice handler and the output from publication_to_biblio

nics (approved):

PR #870, opened 2022-11-30
fix issue 850

### LINKED ISSUE #850: Allow "select all" and "deselect all" in filters with checkbox options
# UX crash

You can now only click "select all" – which is blocking when you've accidentally checked 50 filter boxes or want to work quickly. Allow "select all" and "deselect all" from filters button.

# Context

## Steps to reproduce

1. Open filter with checkbox options, such as "year"
Image
2. Click "select all"
Image
3. Selecting one year now requires you to
  - A: refresh the page, which resets all filters, and you have to completely start over with your filter set.
  - B: uncheck each box you accidentally activated, which can take ages as well.

Image

_Note; you could argue clicking next to the box without hitting "apply" and clearing the list, but this pattern is also counter-productive. See #844._ 

## Expected behaviour

Change the button to "deselect all"

Image

nicolasfranck:
fixes #850 

@netsensei not sure what to do with that package-lock.json
that constantly seems to change. I've included it now, but
probably not a good idea?


nics:
@miet @nicolasfranck Is this ready to be merged or still some discussion around it?

nics (approved):

PR #884, opened 2022-12-07
auto focus first form-control on htmx load

### LINKED ISSUE #883: When adding an author, automatically put the cursor in the "first name" field
# Enhancement

When adding an author, automatically put the cursor in the "first name" field.

Context: when librarians add people, it is an extra action to undertake and slows them down unnecessarily. If their cursor would already be in "first name" they can always use the tab to got "last name" if needed.

## Current behaviour

1. Go to any publication or dataset, and navigate to People & Affiliations
2. Click "Add (author, editor, supervisor, ...)"
3. To start searching for a person, first click on "first name"
Image
4. Start typing
Image

## Expected behaviour
1. Go to any publication or dataset, and navigate to People & Affiliations
2. Click "Add (author, editor, supervisor, ...)"
3. Immediately be able to start typing in the "first name" field
Image

nicolasfranck:
fixes #883


nics:
We probably want a more targeted class like 'form-autofocus' to have more control over this behaviour

nicolasfranck:
Updated: now it only focus on first element with class `form-control-auto-focus`.

The template for `contributor_name` has now that class, so it has effect on adding authors, supervisors and other types of contributors

nics (approved):

PR #885, opened 2022-12-07
Issue 881: add transfer command

netsensei:


netsensei:
@nics Re-implemented the transfer command.

nics (approved):
looking' good

PR #922, opened 2022-12-14
make incremental indexing for frontoffice better

nicolasfranck:


nicolasfranck:
Depends on https://github.ugent.be/Universiteitsbibliotheek/biblio/pull/56

nicolasfranck:
Done:

* added field `date_from` that contains the value from the same same, but formatted in Zulu time. Useful for biblio indexation, as it contains more information, and can be used as a value to store in a timestamp file. To be ignored for indexation of course in the frontend.
* fields `date_created` and `date_updated` should be UTC. In the old librecat these were stored in zulu time as strings, and biblio just splitted them; so there is an explicit assumption that "2010-01-01 11:00" is that time in UTC. Old records that originate from before the migration have zulu time in those fields (and will already be correct), but newer records have local time with time zone appended, which leads to some confusion, especially if you lose the time zone in biblio.
* filter `updated_since` now filters on `date_from` instead of on `data->>"date_updated"`. This has a double advantage: you are querying on a non json field (faster), and the format of the field is consistent. The value of field date_from is supposed to contain the same value as data.date_updated anyway

PR #925, opened 2022-12-14
UserService/PersonService/ProjectService: use backend mongodb directly

### LINKED ISSUE #923: Connect to mongodb directly for retrieval of logged in users
Currently the backoffice uses a handler in the frontoffice when collecting user information,
which slows down the backoffice. The backoffice should connect directly to mongodb, the
action that biblio does now.

nicolasfranck:
Fixes #923 

Done:

* Implemented alternate `UserService` that directly connects to authority database in mongodb
* Implemented alternate `PersonService` that directly connects to authority database in mongodb
* Implemented alternate `ProjectService` that directly connects to authority database in mongodb
* Added viper argument `mongodb-uri`

TODO:

* Add environment variable `BIBLIO_BACKEND_MONGODB_URI` and set it to something like `mongodb://user:password@localhost:27017/?maxPoolSize=20&w=majority&connectTimeoutMS=1000&serverSelectionTimeoutMS=1000`
* check [mongodb connections](https://www.mongodb.com/docs/drivers/go/current/fundamentals/connection/#connection-options) as the defaults allow for too large timeouts (which leaves the websites hanging). I noticed that `connectTimeoutMS` and `serverSelectionTimeoutMS` are set to 30s by default..

Notes:

* no need for reconnection to mongodb. Is done automatically


nics:
@nicolasfranck great thanks
is this ready to be merged? it has a TODO => you can make the timeouts smaller
btw did you check if the nomad machines have access to mauriac/mauriac9/mauriac0?

nicolasfranck:
@nics access to the mongo cluster needs to be checked indeed, and variable needs to be added. I guess we control those variables, right?

nics:
Dependent on firewall access: https://redmine.inuits.eu/issues/104646

nicolasfranck:
I have added the missing (search) services that were still part of the biblioClient.

This requires add additional environment variable `BIBLIO_BACKEND_FRONTEND_ES6_URL`,
which has the exact same syntax as `BIBLIO_BACKEND_ES6_URL`. Not sure if it is necessary
to specify a separate source url (I suspect backend and frontend ES may be located else where in the future?)

PR #983, opened 2023-01-18
File upload show progress

nicolasfranck:
Done:

* show upload progress while uploading (single) file
* on error show dialog. Close button is link that reloads the current page (and tab)
* fix weird behaviour when uploading file bigger than 2G (the file size limit): server receives file three times, while at the client level this is not visible. Even worse, the client reports an empty response. Investigation revealed that the server does send a response, but the client doesn't read it because it is still uploading its body. Fixed by fully reading the request body, to `io.Discard` if needed. Only if the upload file size is small enough, the client will likely have uploaded everything when it received the response. Typical for streaming servers where the uploaded file is not a temporary file? cf. https://stackoverflow.com/questions/43777433/go-http-maxbytesreader-resetting-connection

TODO: 

* @mietcls current error when file upload exceeds the limit is somewhat "vague". "file upload failed"
* make "disabled" state more visible. This has been implemented in the style (see https://biblio-backend-ugent-prototype-git-master-ugent-library.vercel.app/styleguide/c-file-upload.html), but cannot be used yet. @nics ?


nics:
@nicolasfranck upload progress part replaced by 006910f99a27f14b070b9a3f32ba730202bbf4a7

nics:
@nicolasfranck did you see the question above?

nicolasfranck:
@nics of course, but it is a bit messy now to overlook what changed and how to re-fix the merge conflict

PR #1014, opened 2023-02-08
json marshalling of nil slices results into "null"

nicolasfranck:


nics (approved):
I'm going to merge this to fix the bug, but form.Text really shouldn't contain a json string, just make another form field type (e.g. with an extra hidden field)

PR #1102, opened 2023-05-12
Feature/escleanup

netsensei:
* Moves indexing listeners to `Services()`
* Uses bulk indexers for everywhere, doesn't use sync indexing.
* Fixes integration tests for arm64


netsensei:
Still needs `Cobra.OnFinalize()`. Implementing this now.

nics (approved):

PR #1106, opened 2023-05-25
Refine licenses for datasets

mietcls:
@nicolasfranck does this change also immediately update the OAI? Or is there an extra step needed? If so, can we take an extra look together?

### DIFF
--- a/internal/app/handlers/frontoffice/handler.go
+++ b/internal/app/handlers/frontoffice/handler.go
@@ -28,16 +28,16 @@ import (
 const timestampFmt = "2006-01-02 15:04:05"
 
 var licenses = map[string]string{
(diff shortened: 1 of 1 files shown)


nicolasfranck:
@mietcls it does not change `<dcterms:license>` as that always contains the "code", but it will change the copyright statement in `<accessCondition lang="eng" type="useAndReproduction">`

mietcls:
Is something like this an option?

```
var licenses = map[string]string{
	"publication_licenses.CC0-1.0":         "Creative Commons Public Domain Dedication (CC0 1.0)",
	"dataset_licenses.CC0-1.0":               "Creative Commons Zero v1.0 Universal",
}
```

```
var openLicenses = map[string]struct{}{
	"publication_licenses.CC0-1.0":     {},
	"dataset_licenses.CC-BY-4.0":      {},
}
```

PR #1113, opened 2023-06-01
only save changes for publications/datasets when changes present

nicolasfranck:
* `SavePublication/SaveDataset`: handles both new records and preexisting records, so an old record must not preexist here. If there are no changes, saving is canceled.
* `UpdatePublication/UpdateDataset`: handles only preexisting records, so old record must exist. If there are no changes, saving is canceled.


nics:
@nicolasfranck  i think your approach is bit too simplistic. If you feed it an invalid record, it will return nil (= all ok).

Also at this stage it's time to be stricter about style and readability, e.g.

```
func (s *Repository) SaveDataset(d *models.Dataset, u *models.User) error {
	old, err := s.GetDataset(d.ID)
	if err != nil && err != backends.ErrNotFound {
		return err
	}
	if reflect.DeepEqual(old, d) {
		return nil
	}
```

instead of:

```
func (s *Repository) SaveDataset(d *models.Dataset, u *models.User) error {

	if oldDataset, err := s.GetDataset(d.ID); err != nil {
		if err != backends.ErrNotFound {
			return err
		}
	} else {
		if reflect.DeepEqual(oldDataset, d) {
			return nil
		}
	}
```

nics:
@nicolasfranck validation rules change
i think we have to be careful not to change the existing behaviour

nicolasfranck:
@nics yes but you cannot fix that here in a method that does not do validation.

nics (approved):

PR #1157, opened 2023-07-13
use pubmed search arg resulType=core and use richer attributes

### LINKED ISSUE #1153: Abstract of Pubmed is not included (also does not work in prod)

nicolasfranck:
Fixes #1153 , and adds other attributes from pubmed api, using the more rich api:

* add abstract. Language is always set to `eng`. ok?
* uses array `authorList.author` which holds an array of objects, instead of trying to parse `authorString`. Advantage is that `firstName` and `lastName` can now be extracted in full, while previously it showed `<first-name> <abbreviation-of-last-name`. e.g. `Gonzalez N`
* retrieve full journal title instead of abbreviation (`Plant physiology` instead of `Plant Physiol`)
* add language (different from language of abstract?)

### DIFF
--- a/internal/backends/pubmed/client.go
+++ b/internal/backends/pubmed/client.go
@@ -102,13 +102,13 @@ func (c *Client) GetPublication(id string) (*models.Publication, error) {
 	}
 	if authorList := attrs.Get("authorList.author"); authorList.IsArray() {
 		for _, author := range authorList.Array() {
-			var firstName string
-			var lastName string
-			if fn := author.Get("firstName"); fn.Exists() {
-				firstName = fn.String()
+			firstName := author.Get("firstName").String()
+			if firstName == "" {
+				firstName = "[missing]"
 			}
-			if ln := author.Get("lastName"); ln.Exists() {
-				lastName = ln.String()
+			lastName := author.Get("lastName").String()
+			if lastName == "" {
+				lastName = "[missing]"
 			}
 			c := models.ContributorFromFirstLastName(firstName, lastName)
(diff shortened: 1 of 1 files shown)


internal/backends/pubmed/client.go
```diff
@@ -13,6 +13,23 @@ import (
 	"github.com/ugent-library/biblio-backoffice/internal/models"
 )
 
+// api reference: https://europepmc.org/RestfulWebService
+/*
```
nics, line 17: thisis not needed

internal/backends/pubmed/client.go
```diff
@@ -78,23 +96,38 @@ func (c *Client) GetPublication(id string) (*models.Publication, error) {
...
 		p.Year = res.String()
 	}
-	if res := attrs.Get("authorString"); res.Exists() {
-		for _, r := range reSplit.Split(res.String(), -1) {
-			nameParts := strings.Split(strings.ReplaceAll(r, ".", ""), " ")
-			firstName := nameParts[len(nameParts)-1]
-			lastName := strings.Join(nameParts[:len(nameParts)-1], " ")
+	if authorList := attrs.Get("authorList.author"); authorList.IsArray() {
+		for _, author := range authorList.Array() {
+			var firstName string
```
nics, line 123: `firstName := author.Get("firstName").String()`

I think the other importers also add `[missing]` if a name part is empty (see crossref importer)

nicolasfranck: done

nics (approved):

Issue #1137, opened 2023-06-14
Provide error: The import service for DOIs cannot be reached (Crossref or DataCite)


mietcls:
# Enhancement
Provide error: The import service for DOIs cannot be reached (Crossref or DataCite)
> Connection errors => CrossRef is unreachable (The wire between the Backoffice & Crossref is blocked for whatever technical reason; or the CrossRef API is offline)

For context, see epic #1091 

## Success criteria

Whenever there are connection issues, and the import service for DOIs can not be reached (Crossref or DataCite), I see an error toaster. See implementation suggestions.

### Publications
![Scherm­afbeelding 2023-06-14 om 17 35 54](https://github.com/ugent-library/biblio-backoffice/assets/1572885/be166555-7f90-4bf5-a6c2-d129b5f77bd9)

To be adapted:
- The route to other import options (now just a link)
- The error code

```
<div class="toast show" role="alert" aria-live="assertive" aria-atomic="true" data-autohide="false">
    <div class="toast-header">
        <div class="u-divide-x">
            <span class="c-subline">DOI import</span>
            <span class="c-body-small text-muted ml-3 pl-3">404</span>
        </div>
        <button class="close" type="button" data-dismiss="toast" aria-label="Close">
            <i class="if if-close"></i>
        </button>
    </div>
    <div class="toast-body">
        <i class="if if-alert-fill if--warning"></i>
        <div class="toast-content">
            <h3 class="alert-title">The publication could not be imported.</h3>
            <p class="mb-3">The import service <a href="https://www.crossref.org" target="_blank" class="text-muted">Crossref</a> could not be reached.</p>
            <p>Try again later, or <a href="https://backoffice.biblio.ugent.be/publication/add">use a different import option</a>.</p>
        </div>
    </div>
</div>
```

### Datasets

![Scherm­afbeelding 2023-06-14 om 17 29 33](https://github.com/ugent-library/biblio-backoffice/assets/1572885/06304691-512c-4ca2-a61b-249671488a79)

To be adapted:
- The route to other import options (now just a link)
- The error code

```
<div class="toast show" role="alert" aria-live="assertive" aria-atomic="true" data-autohide="false">
    <div class="toast-header">
        <div class="u-divide-x">
            <span class="c-subline">DOI import</span>
            <span class="c-body-small text-muted ml-3 pl-3">404</span>
        </div>
        <button class="close" type="button" data-dismiss="toast" aria-label="Close">
[...]

Issue #1136, opened 2023-06-14
Provide error: The API of DataCite or Crossref says there is a problem with accessing the record


mietcls:
# Enhancement
Provide error: The import service for DOIs cannot be reached (Crossref or DataCite)
> API errors => The API returns 4xx or 5xx errors. The former: a record could not be found, or can't be accessed; the latter: the API returns some kind of general system error.

For context, see epic #1091 

## Success criteria
Whenever something is broken within the Crossref or DataCite service, I see an error toaster. See implementation suggestions.

### Publications
![Scherm­afbeelding 2023-06-15 om 14 33 43](https://github.com/ugent-library/biblio-backoffice/assets/1572885/11d0b0a5-cc3f-4ff5-a1d9-c18ef25794e8)

To be adapted:
- The route to other import options (now just a link)
- The error code

```
<div class="toast show" role="alert" aria-live="assertive" aria-atomic="true" data-autohide="false">
    <div class="toast-header">
        <div class="u-divide-x">
            <span class="c-subline">DOI import</span>
            <span class="c-body-small text-muted ml-3 pl-3">404</span>
        </div>
        <button class="close" type="button" data-dismiss="toast" aria-label="Close">
            <i class="if if-close"></i>
        </button>
    </div>
    <div class="toast-body">
        <i class="if if-alert-fill if--warning"></i>
        <div class="toast-content">
            <h3 class="alert-title">The publication could not be imported.</h3>
            <p class="mb-3">The DOI was recognised, but <a href="https://www.crossref.org" target="_blank" class="text-muted">Crossref</a> could not access the record.</p>
            <p>Try again later, or <a href="https://backoffice.biblio.ugent.be/publication/add">use a different import option</a>.</p>
        </div>
    </div>
</div>
```

### Datasets
![Scherm­afbeelding 2023-06-15 om 14 34 16](https://github.com/ugent-library/biblio-backoffice/assets/1572885/96880825-cb2c-450c-ad88-5dfc1732053e)

To be adapted:
- The route to other import options (now just a link)
- The error code

```
<div class="toast show" role="alert" aria-live="assertive" aria-atomic="true" data-autohide="false">
    <div class="toast-header">
        <div class="u-divide-x">
            <span class="c-subline">DOI import</span>
            <span class="c-body-small text-muted ml-3 pl-3">404</span>
        </div>
        <button class="close" type="button" data-dismiss="toast" aria-label="Close">
            <i class="if if-close"></i>
[...]

Issue #1135, opened 2023-06-14
Provide error: JSON not recognised


mietcls:
# Enhancement
Provide error: JSON not recognised
> JSON Validation errors => The Crossref JSON itself is invalid (it's not a JSON string; but garbled data)

For context, see epic #1091 

## Success criteria
Whenever we can resolve the DOI but the JSON is not recognised, I see an error toaster. See implementation suggestions.

### Publications
![Scherm­afbeelding 2023-06-15 om 14 45 22](https://github.com/ugent-library/biblio-backoffice/assets/1572885/e0f69584-a6a2-4c97-9614-c9c354540537)

To be adapted:
- The route to other import options (now just a link)
- The error code

```
<div class="toast show" role="alert" aria-live="assertive" aria-atomic="true" data-autohide="false">
    <div class="toast-header">
        <div class="u-divide-x">
            <span class="c-subline">DOI import</span>
            <span class="c-body-small text-muted ml-3 pl-3">404</span>
        </div>
        <button class="close" type="button" data-dismiss="toast" aria-label="Close">
            <i class="if if-close"></i>
        </button>
    </div>
    <div class="toast-body">
        <i class="if if-alert-fill if--warning"></i>
        <div class="toast-content">
            <h3 class="alert-title">The publication could not be imported.</h3>
            <p class="mb-3">The DOI was recognised, but the metadata could not be retrieved because of a problem with  <a href="https://www.crossref.org" target="_blank" class="text-muted">Crossref</a>'s service.</p>
            <p>Try again later, or <a href="https://backoffice.biblio.ugent.be/publication/add">use a different import option</a>.</p>
        </div>
    </div>
</div>
```

### Datasets

![Scherm­afbeelding 2023-06-15 om 14 45 43](https://github.com/ugent-library/biblio-backoffice/assets/1572885/30a4f54c-d443-47e8-aaa7-3a7f23380c15)

To be adapted:
- The route to other import options (now just a link)
- The error code

```
<div class="toast show" role="alert" aria-live="assertive" aria-atomic="true" data-autohide="false">
    <div class="toast-header">
        <div class="u-divide-x">
            <span class="c-subline">DOI import</span>
            <span class="c-body-small text-muted ml-3 pl-3">404</span>
        </div>
        <button class="close" type="button" data-dismiss="toast" aria-label="Close">
            <i class="if if-close"></i>
        </button>
    </div>
    <div class="toast-body">
[...]

Issue #1091, opened 2023-05-03
Error messages needed for failed record imports


mietcls:
>  I don't have a grasp on everything that can go wrong

I think we have these classes of problems:

* Record Validation errors  => the CrossRef JSON is valid; but fields are empty, or data is missing)
* JSON Validation errors => The CrossRef JSON itself is invalid (it's not a JSON string; but garbled data)
* API errors => The API returns 4xx or 5xx errors. The former: a record could not be found, or can't be accessed; the latter: the API returns some kind of general system error.
* Connection errors => CrossRef is unreachable (The wire between the Backoffice & Crossref is blocked for whatever technical reason; or the CrossRef API is offline)

I've implemented validation errors in #1035 but the other types of errors will yield a more generic "something went wrong" error pop-up. We could be more specific; but I don't think we want to burden researchers with the minutiae of how an API works, no?  We do want some human-friendly message that gives at least a general idea of what went wrong (e.g. "CrossRef could not be reached at this time, try again later", "The record could not be found with CrossRef", "The record could not be retrieved from CrossRef")

So, if we know how to word the errors, it shouldn't be that hard to make nicer error messages.

_Originally posted by @netsensei in https://github.com/ugent-library/biblio-backoffice/issues/1088#issuecomment-1533113247_

nicolasfranck:
@netsensei indeed, we cannot burden reviewers with api problems, and if we are too vague in our error messaging and report `something went wrong`, end users will also report us vague problems. 

Anyway, apart from the error messaging itself, weren't we also thinking about generating an identifier, report that to the end user, and note that in the log message next to the error for faster tracking?

netsensei:
> Anyway, apart from the error messaging itself, weren't we also thinking about generating an identifier, report that to the end user, and note that in the log message next to the error for faster tracking?

1/

All incoming requests get a request ID by default. We can display this ID when the server hits a fault condition and returns an error to the end-user. We currently do this in the context of adding or creating new contributors => when an error happens, you see a modal which includes a request ID and mention to contact the review team.

2/

Tagging a request with an ID and showing an error w/ ID can only be fully achieved when a HTTP request was received successfully, and a HTTP response was successfully delivered to the browser. That doesn't happen here: transmission of the HTTP request itself - the upload of a file - can't be completed e.g. because the connection was lost, or because the server itself became unavailable.

So, in that context, there's no use for generating a request ID and including it into an error message which could be pushed to the end user.

[...]

PR #1163, opened 2023-07-26
provide better error messages when importing publications/datasets

### LINKED ISSUE #1138: Provide error: DOI not recognised by trusted service

nicolasfranck:
fixes #1138 #1135 #1136 #1137

So

* provides error for "bad connection"
* provides error for "not found" (if it can be determined)
* provides error for "invalid content"

TODO: I have put the text in translations.go, but .. that contains also html data. So feels a bit off.


mietcls:
@nicolasfranck which parts do you mean? Do we need to rewrite information?

nicolasfranck:
@mietcls not sure if that is possible.

What I mean: for each situation (bad connection, resource not found ..), a localized message (stored in the translation map) needs to be put in the "body" of the toaster.
In the past that body of the toaster was merely a sentence, with here and there simple tags like `<a>` or `<i>`. But now the body needs to have a whole html structure. That fills the translation map with data like `"<div style=\"display:none\">I do not want you to see this</div> are we good?"`, data you would expect in a template...

nicolasfranck:
@mietcls yes it works, but putting a lot of html in a translation map makes it unreadable.

mietcls:
@nicolasfranck yes that's true. There are plans to move this to a different format, where it becomes more manageable.

nicolasfranck:
to delete or split up? @netsensei @nics

netsensei:
@nicolasfranck @nics @mietcls I'm closing this. This an attempt to create error messages which were not useful enough for end users. See #1091  => needs to be turned into a discussion to hash out how we want to move this one forward.

PR #1234, opened 2023-09-21
Feature/bootstrap-5

mietcls:
Add Bootstrap 5. Some items will be broken.

To do:
- Fix JS for **popover** and **tooltips**. You can find the changes here:
https://docs.google.com/document/d/1oWP2xyLvWN0dXTAyT-yqZYzhU-ndes4vRKgSbl0JU_M/edit#heading=h.surjhwoxq9vm


internal/backends/mods36/encoder.go
```diff
@@ -162,9 +162,9 @@ func (e *Encoder) EncodePublication(p *models.Publication) ([]byte, error) {
 	for _, c := range p.Author {
 		addContributor(r, "editor", c)
 	}
-	// TODO use promoter terminology everywhere?
+
 	for _, c := range p.Supervisor {
-		addContributor(r, "promoter", c)
+		addContributor(r, "Supervisor", c)
```
nics, line 167: This exporter is using the marcrelator vocabulary which uses promoter as a term, be careful when replacing terminology that is not user facing.

nics (changes requested):

internal/backends/mods36/encoder.go
mietcls: Thanks, I'll be more paranoid about it!

nics (approved):

PR #1235, opened 2023-09-22
remove dependency on viper in cli (not the client cli)

nics:
- introduces a config struct
- removes the dependency on viper in the cli
- aligns the cli with our other apps
- cleans up some code and old comments
- leaves the client cli alone for now


netsensei (approved):
Reviewed this PR. Looks good to me.

PR #1266, opened 2023-10-17
Add batch method `external_fields.set`

nicolasfranck:
Add batch method `external_fields.set`, and field `ExternalFields` to model Publication.

Needed to add extra enrichment fields (e.g. `ECOOM`).

The field `ExternalFields` is defined as `map[string][]string`, and therefore expects
a kind of flat structure of extra keys, with multiple values.

Example for ECOOM data:

```
{
  "external_fields": {
    "ecoom-bof-css": ["1"],
    "ecoom-bof-sector": ["sector1","sector2"],
    "ecoom-bof-weight": ["2"]
  }
}
```

Batch method `external_fields.set` to be used as such:

```
<id>,external_fields.set,"ecoom-bof-css","1"
<id>,external_fields.set,"ecoom-bof-weight","1"
<id>,external_fields.set,"ecoom-bof-sector","sector1","sector2"
```


nics (commented):
@nicolasfranck models.Identifiers uses the same data structure, maybe define a type like 'models.Values' (cfr url.Values which has the same structure) that is used by both fields

nics (approved):

PR #1273, opened 2023-10-19
add missing enrichment for person data in frontoffice record

nicolasfranck:
In order to deprecate the [catmandu fix in biblio](https://github.ugent.be/Universiteitsbibliotheek/biblio/blob/master/lib/Catmandu/Fix/add_person_data.pm) I added the following to the frontoffice record:

To struct `frontoffice.Person`:

* field `biblio_id` (which has same value as `id`)
* field `name_last_first`

To struct `frontoffice.Record`:

* field `first_author`
* field `last_author`
* field `sole_author`
* field `author_sort`

Field `created_by` is now also filled with all person data needed.

I also changed something in the mapping from the mongodb authority record of a person:

* when a person has no `ugent_department_id`, but `object_class` contains `ugentFormerEmployee`, then affiliation `UGent` is added
* when `object_class` contains `uzEmployee`, then affiliation `UZGent` is added. This has as result that in the contributor list the "ugent icon" is visible below column "affiliation", with a blue label next to it having "UZGent".

### DIFF
--- a/frontoffice/record.go
+++ b/frontoffice/record.go
@@ -53,7 +53,7 @@ type Affiliation struct {
 
 type Person struct {
 	ID            string        `json:"_id,omitempty"`
-	BIBLIOID      string        `json:"biblio_id,omitempty"`
+	BiblioID      string        `json:"biblio_id,omitempty"`
 	CreditRole    []string      `json:"credit_role,omitempty"`
 	Name          string        `json:"name,omitempty"`
 	FirstName     string        `json:"first_name,omitempty"`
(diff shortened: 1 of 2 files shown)


frontoffice/record.go
```diff
@@ -52,14 +52,16 @@ type Affiliation struct {
...
-	ID          string        `json:"_id,omitempty"`
-	CreditRole  []string      `json:"credit_role,omitempty"`
-	Name        string        `json:"name,omitempty"`
-	FirstName   string        `json:"first_name,omitempty"`
-	LastName    string        `json:"last_name,omitempty"`
-	UGentID     []string      `json:"ugent_id,omitempty"`
-	ORCID       string        `json:"orcid_id,omitempty"`
-	Affiliation []Affiliation `json:"affiliation,omitempty"`
+	ID            string        `json:"_id,omitempty"`
+	BIBLIOID      string        `json:"biblio_id,omitempty"`
```
nics, line 56: @nicolasfranck Biblio is not an acronym so should be BiblioID

nics (approved):

PR #1322, opened 2023-11-22
Integrate project suggestions box with the projects service

netsensei:


netsensei:
@nics Okay, this should be ready for review / merge.

* Integration with OpenAPI client from project-service
* Updated templates by @mietcls 

To discuss / do in a follow up issue: how do we handle (unsafe) html tags which are part of the description / abstract field.

backends/projects/client.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	return &Client{
+		config: c,
+		client: client,
+	}, nil
+}
+
+func (c *Client) GetProject(id string) (*models.Project, error) {
+	ctx := context.Background()
```
nics, line 43: better to use context.TODO() if it's a placeholder

backends/projects/client.go
```diff
@@ -0,0 +1,148 @@
...
+		p := c.mapper(&hit)
+		projects = append(projects, *p)
+	}
+
+	return projects, nil
+}
+
+func (c *Client) mapper(ap *api.GetProject) *models.Project {
+	p := &models.Project{
+		EUProject: &models.EUProject{},
```
nics, line 80: Shouldn't EUProject be nil if it isn't a eu project?

cli/config.go
```diff
@@ -78,6 +78,10 @@ type Config struct {
 		APIURL string `env:"API_URL"`
 		APIKey string `env:"API_KEY"`
 	} `envPrefix:"OAI_"`
+	Projects struct {
+		APIURL string `env:"API_URL"`
+		APIKEY string `env:"API_KEY"`
```
nics, line 83: better use `APIKey` for consistency (see lines above)

mietcls:
@netsensei We've got the same issue with abstracts when importing them for datasets from other sources.

There's a conversation going on here about parsing links, that might be an opportunity to create something reusable? https://github.com/ugent-library/biblio-backoffice/discussions/1324

backends/projects/client.go
netsensei: Done.

cli/config.go
netsensei: Done.

backends/projects/client.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (c *Client) SuggestProjects(q string) ([]models.Project, error) {
+	ctx := context.Background()
+
+	res, err := c.client.SuggestProjects(ctx, &api.SuggestProjectsRequest{Query: q})
+	if err != nil {
+		return nil, err
+	}
+
+	projects := make([]models.Project, 0)
```
nics, line 68: this better written as:

```
        projects := make([]models.Project, len(res.Data))

	for i, hit := range res.Data {
		p := c.mapper(&hit)
		projects[i] = *p
	}
```

much more efficient

you can gain an extra line if you omit the temporary `p` variable

netsensei: done. temp var is a pointer variable; hard to omit since we can't directly dereference a function pointer return.

PR #1333, opened 2023-11-29
Feature/card structure

mietcls:
**Question:**
- Added the "copy" button the way it was implemented in Deliver – but it needs a different approach here. Can someone review / give suggestions?

Done:
- Starting to align cards with Emily's new design that will make us ready for Plato import additions at some point.
- Added some long-standing requests from reviewers.

Later:
- After this PR a ["show/hide details" button](https://biblio-backend-ugent-prototype-git-bootstr-307d84-ugent-library.vercel.app/plato-test/overview.html) needs to be added to the card so the layout can be finished
  - Projects can be shown with the show-hide button
  - Creation date added can be shown with the show-hide button
- Try to align copy component with Deliver visually (table issue)


nics (changes requested):

views/publication/search_hit.gohtml
```diff
@@ -1,111 +1,143 @@
...
+            {{else}}
+                {{if .Publication.DOI}}
+                <a class="dropdown-item" href="https://www.altmetric.com/explorer/outputs?scope=all&identifier={{.Publication.DOI}}" target="_blank">
+                    <i class="if if-external-link"></i>
+                    <span>Altmetric</span>
+                </a>
                 {{end}}
-            </div>
+            {{end}}
+            {{/*
```
nicolasfranck, line 51: Best to leave out code in comment. Probably copied and pasted from another source? This block runs from line 52 till line 63

nicolasfranck: never mind, was already in the block on the left, but looked a little different though

mietcls:
I need to review this one entirely after a conversation with Emily.

views/publication/search_hit.gohtml
mietcls: Can I completely drop this? Or was there something else that doesn't look right? Thanks for the review.

Issue #851, opened 2022-11-23
Allow reviewers and curators to withdraw records even when it is on locked


mietcls:
# UX crash

Allow librarians to withdraw records even when it is on locked.

Sometimes librarians want to protect researchers from editing information. However, when they need to edit the publication type, they cannot because it is on locked. This means they have to unlock, confirm they want to unlock, withdraw, edit publication type, republish, lock, confirm they want to lock again. It is a superfluous block.

## Current situation

Cannot withdraw when the record is on "locked"
![Image](https://user-images.githubusercontent.com/1572885/203584472-948dccc5-574b-4b5d-904a-95a0ff1240ad.png)

Can only withdraw when unlocked
![Image](https://user-images.githubusercontent.com/1572885/203584642-2004a6f7-5754-4355-8cda-efab68678cb8.png)

## Expected behaviour

Allow librarians to withdraw records even when it is on locked.
![Image](https://user-images.githubusercontent.com/1572885/203584893-9b64c69f-1c86-4d45-913c-7989ec4cb5ad.png)

nicolasfranck:
@mietcls well, they can, if they first unlock the record.

If locking is to be ignored, we also have to ignore that flag in other situations, not?
Like editing of data.

Of course, they want to be able to work something without the risk that
someone else is interfering?

nicolasfranck:
anyway, technically already possible, just a missing button

mietcls:
@nicolasfranck the issue states that it should be possible *during* unlock because it interrupts their workflow unnecessarily. So the button should just be there indeed.

Other workflows or functionalities do not need to be touched. They can already edit documents when they are on locked, it's only to keep researchers from editing further.

mietcls:
@nicolasfranck are you picking this one up or are you merely commenting?

nicolasfranck:
@mietcls merely commenting

PR #1334, opened 2023-12-01
Always show withdraw button for librarians

mietcls:
Should fix https://github.com/ugent-library/biblio-backoffice/issues/851

Shows withdraw button for librarians even when locked
- "view as" does not work locally; so not sure if this works as expected for researchers. The withdrawn button should _not_ show up for researchers when the publication or dataset is locked.
- Nested it in a dumb way and could use some more simple if-else love, but got stuck along the way. Any tips?


nicolasfranck:
@mietcls I have changed some things, i.e. added methods like `CanPublishPublication` and  `CanWithdrawPublication` to make these actions more readable. Of course `CanRepublishPublication` sounds a bit superfluous, and it does basically the same thing internally.

PR #1353, opened 2024-01-11
oai review

nicolasfranck:
* reviewed oai_dc and mods_36 conversion in comparison to old conversion
* added required field `MetadataPrefix` to delete record request, and did this for all inserted formats. See also change in oai-service that retroactively inserts records with this method


nicolasfranck:
replaced by https://github.com/ugent-library/biblio-backoffice/pull/1358

PR #1374, opened 2024-02-14
Devcontainer fixes

nicolasfranck:
Added some minor changes for devcontainers:

* removed `user: postgres` from docker-compose.yml so that postgres can start up correctly, and does not try to set the user of the folder
* removed non existing example command for seeding from Dockerfile (misleading)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -27,7 +27,7 @@ Here are some useful commands to get you started:\n\
 Start the app:\n\
     reflex -d none -c reflex.docker.conf\n\
 \n\
-Purge all data from the database:\n\
+Purge all data from the database and initialize all tables/indexes:\n\
     go run main.go reset --force\n\
 \n\
 Happy hacking!\n"\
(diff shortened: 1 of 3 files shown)


.devcontainer/Dockerfile
```diff
@@ -27,14 +27,8 @@ Here are some useful commands to get you started:\n\
...
     reflex -d none -c reflex.docker.conf\n\
 \n\
-Seed the app with dummy data:\n\
-    go run main.go seed\n\
-\n\
-If you already have data in the database, you can force seeding:\n\
-    go run main.go seed --force\n\
-\n\
 Purge all data from the database:\n\
-    go run main.go reset --force\n\
```
netsensei, line 37: Leave it to `--force`.  We actually want to adhere to common conventions as to how we name commonly used switches.  e.g. https://clig.dev/

Create a issue with a PR renaming this switch instead in the command to `force`.

netsensei (changes requested):

.devcontainer/Dockerfile
nicolasfranck: was probably a typo. Not intentionally

nicolasfranck: agreed. I have replaced `--confirm` by `--force` in the reset command (like in deliver), and added extra documentation to the example reset command here, so people know what it also does (like making sure that the indexes exist)

PR #1376, opened 2024-02-16
fix issue 1375

### LINKED ISSUE #1375: Show clickable urls and newlines in reviewer comments
# Enhancement

Comments that are entered by reviewers in the publication/dataset tab `Biblio messages`
are shown in the list of datasets/publications. These comments often contain url's and
newlines. 

Right now these urls are shown as they are (i.e. plain text), and therefore require an extra
copy and paste to visit them.

Also newlines have no meaning in html, so are made useless.

## Expected behavior
In order to make the visualisation better:

* convert every url in the plain text block into a clickable url. Important: make sure that url's on the end of a line, or within brackets for example are also recognized.
* convert every newline into an `<br>`

nicolasfranck:
fixes #1375 

how:

* added new func `renderMsg` to template funcmap that receives (unsafe) string and returns html safe string (template.HTML). Implemented as a filter.
* `renderMsg` does the following:
  * html escape input
  * replace newlines by `<br>`
  * replace all https links by `<a>`. Extracting urls was not trivial, as in all of the examples, the urls where hardly ever put between spaces, but mostly between parenthesises, before a dot, after a colon.. So I could not use something like "what is parseable as an url", as that required word extraction first. The limitation on https is due to the library that either allows for matching on a whole set of [schemes](https://pkg.go.dev/mvdan.cc/xurls/v2#Strict) (from http to file, which to much), or either one [scheme](https://pkg.go.dev/mvdan.cc/xurls/v2#StrictMatchingScheme). I think https is enough? Didn't feel the urge to search any further.

### DIFF
--- a/urls/funcmap.go
+++ b/urls/funcmap.go
@@ -94,8 +94,7 @@ func queryClear(u *url.URL) (*url.URL, error) {
 func renderMsg(oldVal string) template.HTML {
 	oldVal = html.EscapeString(oldVal)
 
-	oldVal = strings.ReplaceAll(oldVal, "\r\n", "\n")
-	oldVal = strings.ReplaceAll(oldVal, "\n", "<br>")
+	oldVal = strings.ReplaceAll(oldVal, "\r\n", "<br>")
 
 	re, _ := xurls.StrictMatchingScheme("https")
 	urlIndexPairs := re.FindAllStringIndex(oldVal, -1)
@@ -112,7 +111,7 @@ func renderMsg(oldVal string) template.HTML {
 		//URL
 		postfix := oldVal[pair[0]:pair[1]]
 		newValBuilder.WriteString("<a href=\"")
-		newValBuilder.WriteString(html.EscapeString(postfix))
+		newValBuilder.WriteString(postfix)
 		newValBuilder.WriteString("\" target=\"_blank\">")
 		newValBuilder.WriteString(postfix)
 		newValBuilder.WriteString("</a>")
--- a/urls/funcmap.go
+++ b/urls/funcmap.go
@@ -1,11 +1,14 @@
 package urls
 
 import (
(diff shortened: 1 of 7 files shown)


urls/funcmap.go
```diff
@@ -86,3 +90,38 @@ func queryClear(u *url.URL) (*url.URL, error) {
...
+		// NON URL
+		prefix := oldVal[startPos:pair[0]]
+		if len(prefix) > 0 {
+			newValBuilder.WriteString(prefix)
+		}
+
+		//URL
+		postfix := oldVal[pair[0]:pair[1]]
+		newValBuilder.WriteString("<a href=\"")
+		newValBuilder.WriteString(html.EscapeString(postfix))
```
nics, line 115: Why is this double escaped?

urls/funcmap.go
```diff
@@ -86,3 +90,38 @@ func queryClear(u *url.URL) (*url.URL, error) {
 
 	return &newU, nil
 }
+
+func renderMsg(oldVal string) template.HTML {
+	oldVal = html.EscapeString(oldVal)
+
+	oldVal = strings.ReplaceAll(oldVal, "\r\n", "\n")
```
nics, line 97: Why is this necessary?

urls/funcmap.go
```diff
@@ -86,3 +90,38 @@ func queryClear(u *url.URL) (*url.URL, error) {
...
+	oldVal = strings.ReplaceAll(oldVal, "\n", "<br>")
```
nics, line 98: Why is this necessary?

urls/funcmap.go
nicolasfranck: I'm just not sure that all newlines are windows/internet newlines (where `\r\n` is used). Better safe than sorry?

PR #1384, opened 2024-03-04
Fix/cypress testing

mietcls:
Fixes testing for Cypress and #1386.
Remaining failing test is from that bug https://github.com/ugent-library/biblio-backoffice/issues/1386, because cypress still tests the dev branch – this will be fixed later.


cypress/support/commands/extract-biblio-id.ts
```diff
@@ -11,7 +11,7 @@ export default function extractBiblioId(subject: undefined | JQuery<HTMLElement>
 
     cySubject = cy.wrap(subject, { log: false })
   } else {
-    cySubject = cy.get('.list-group-item-main', { log: false })
+    cySubject = cy.get('.c-thumbnail-and-text', { log: false })
```
verheyenkoen, line 14: I would replace this with this selector: `.card > .card-body > .list-group > .list-group-item`. Did some quick analysis and that should work during import and in the overview for selecting the individual publications/datasets.

verheyenkoen:
Remaining failing test is from an unrelated bug (issue yet to be created).

mietcls:
Should be fixed now

verheyenkoen:
> I'll create a new PR to fix the bug, then it will pass. However, should this block the PR? This workflow might need a change.

I don't think the failing tests blocks merging the PR. The button is not disabled anyway.

mietcls:
Okay, I will not adapt the PR in that case. We can solve both in this one.

verheyenkoen (approved):

nics (approved):

Issue #961, opened 2023-01-06
[filters] Prioritise filter sequence and visibility


mietcls:
# Story

Solution to be hashed out together.

Part of #954 

Prioritise filter sequence and visibility for librarians.

We have a lot of filters. Not each of them is just as important, and due to the cognitive load we find librarians looking for the right filter again and again.

<img width="1262" alt="Scherm­afbeelding 2023-01-06 om 16 59 02" src="https://user-images.githubusercontent.com/1572885/211049404-cd6b62c5-f22c-40cd-aa27-e1695a717716.png">

## Proposed solution for publications reviewer view

Solution: multiline & drawer

Success criteria
- [x] Put the most important filters first, secondary filters second
  - First line: Biblio status, Classification, Faculty, Publication year, Publication type
  - Second line: Publication status, Librarian tags, Message, Locked, UGent
  - Third line (hidden): WoS type, VABB type, File, File type, Created since, Updated since, Legacy
- [x] Hide filters that are rarely used behind a "show more" button, in the third line
_FYI: there could be a fourth line in the future if we add filters_
- [x] Make sure active filters are always visible.
_If we would select a rarely used filter from the "show more" drawer, and then close the drawer again, it should still be visible on top._
- [x] This is for librarians only in publications

Only important filters shown:
<img width="1280" alt="211817794-a8a37406-6a1e-42b8-85f4-b822b65373a8-b" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/dc807064-98d2-45da-b259-64c6ffa149d5">

All filters shown:
<img width="1278" alt="211823172-34f0916d-2bf5-4062-a922-a3c362755d74-b" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/eb70986c-e7ed-43ce-8ca6-db307e233c35">

Only important filters shown, but with an active rarely used filter option:
<img width="1279" alt="211818153-b8587ecb-3904-49ec-b0dd-f424a5419a63-b" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/b0960b1d-99ab-4743-bead-d19069fb85c4">

## Proposed solution for datasets reviewer view, and researchers view for datasets and publications

Solution: multiline (no drawer)

Success criteria
- [x] Put the most important filters first, secondary filters second
  - [x] Datasets reviewer view
    - First line: Biblio status, Faculty, Locked, Persistent identifier type
    - Second line: Librarian tags, Message, Created since, Updated since
[...]

PR #1392, opened 2024-03-12
Prioritise filter sequence and visibility #961

verheyenkoen:
This PR adds UX features to facet filters for curators in the publications view as described in #961.


models/search_args.go
```diff
@@ -1,12 +1,21 @@
...
-	Sort     []string            `query:"sort,omitempty"`
-	PageSize int                 `query:"page-size"`
-	Facets   []string            `query:"-"`
+	Query      string              `query:"q,omitempty"`
+	Filters    map[string][]string `query:"f,omitempty"`
+	Page       int                 `query:"page"`
+	Sort       []string            `query:"sort,omitempty"`
+	PageSize   int                 `query:"page-size"`
+	Facets     []string            `query:"-"`
+	FacetLines [][]string          `query:"-"`
```
verheyenkoen, line 18: Not sure if this `query:"-"` annotation at the end is right here (or what it does).

views/search/_facets.gohtml
```diff
@@ -1,27 +1,64 @@
...
-                        <i class="if if-close"></i>
-                        <span class="btn-text">Reset filters</span>
-                    </button>
-                </form>
-            </div>
+  </div>
+  <div class="bc-toolbar-right">
+    <div class="bc-toolbar-item">
+      <div class="c-button-toolbar">
+        {{if gt (len .SearchArgs.FacetLines) 2}}
```
verheyenkoen, line 20: I created a constant const for this number of lines (2) [here](https://github.com/ugent-library/biblio-backoffice/blob/7f30bcb376ef31bb76d37533af9f866a06c3c422/models/search_args.go#L9) but not sure how to use that in gohtml. Probably would need to pass that on with the context?

verheyenkoen:
@mietcls Got a working version that you can maybe already have a look at and provide feedback for. For now I'll leave this draft but it is testable if you want.

nics (approved):

views/search/_facets.gohtml
nics: we clean that up after the templ rewrite

Issue #1370, opened 2024-02-07
Make created, edited and system update timestamp more informative.


mietcls:
# Enhancement
Make created, edited and system update timestamp more informative.

Please use the new card layout, if it is not in main yet, please use dev.

## Expected behaviour
Right now the text says "2 months ago". This is not helpful to researchers nor reviewers. They need to be able to see validation of their actions and be able to figure out when something went wrong.

Current:
![Screenshot 2024-02-07 at 10 58 15](https://github.com/ugent-library/biblio-backoffice/assets/1572885/97186f2b-4e8a-4c72-acfe-a67e6722f674)

Expected:
![Screenshot 2024-02-07 at 10 54 11](https://github.com/ugent-library/biblio-backoffice/assets/1572885/39bbd464-194a-4e1b-b735-34ddcd2f495f)
- Created DD/MM/YYYY hh:mm by _name_
- System update DD/MM/YYYY hh:mm
- Edited DD/MM/YYYY hh:mm by _name_

## Context
- Related improvements: 
  - https://github.com/ugent-library/biblio-backoffice/issues/1365
  - https://github.com/ugent-library/biblio-backoffice/issues/1364
- Which timezone to use: https://github.com/ugent-library/biblio-backoffice/issues/935

# Automatic testing scenario
Not sure how to test this one, but seems useful @verheyenkoen?

verheyenkoen:
FYI: For publications/datasets with a system update (as last edit), you don't see the last human edited time but this:

> Created 18/03/2024 10:40 by Biblio Researcher. System update 18/03/2024 11:01. Last edit by Biblio Librarian.

If the last edit was a human edit, you see this:

> Created 18/03/2024 10:48 by Biblio Researcher. Edited 18/03/2024 10:48 by Biblio Researcher.

verheyenkoen:
Update: format should be "YYYY-MM-DD HH:mm", not "DD/MM/YYYY HH:mm".

Issue #935, opened 2022-12-20
Uploaded time of files is off


mietcls:
# Bug

Uploaded time of files is off

TODO:  create separate time epic

## Steps to reproduce

1. Go to Biblio
2. Go to full text & files
3. Look at the time
4. Upload a file
5. See the time is off by one hour

![Image](https://user-images.githubusercontent.com/1572885/208708308-57b143fb-1378-467a-8aba-618bd936d932.png)

![Image](https://user-images.githubusercontent.com/1572885/208708287-85b6fbaa-cff5-4e67-8828-40e9ea90d826.png)

nicolasfranck:
I can confirm this. Timestamp is created correctly in UTC timezone, but shown incorrectly (it should convert from UTC to local time)

mietcls:
@nicolasfranck It's still off unfortunately, it was 17:13 on my machine

<img width="1017" alt="Scherm­afbeelding 2023-02-07 om 17 13 57" src="https://user-images.githubusercontent.com/1572885/217300435-5cf6ff0b-69dd-42e7-8c87-ae994d54d5c9.png">

nicolasfranck:
Same here: https://backoffice.bibliotest.ugent.be/publication/01GRRB1X35W7WE5FB7W8KQKX9J?redirect-url=%2Fpublication&show=files

file was upload at 13:56, but it shows 12:56. Apparently "local" time means something different on the server. Will try to force it to "Europe/Brussels"

nics:
@mietcls what seems most intuitive to you?

- show utc time with timezone indication
- show brussels time
- show local time of the user (extract timezone from request)

nicolasfranck:
I would always add timezone information at the end. 
Just keeping utc as our default would just anger 99% of our users, not? They would always have to do the math.

mietcls:
@nics showing the local time of the user is the best option, as we don't know where they are. People usually point it out when they notice the hour is not the same on the device they are using. Devices use the local time as well.

mietcls:
I'm also going to create an issue to fix the date notation for records later – the "9 months ago" is not useful for the reviewers nor researchers at this point (oldest record we can find)

mietcls:
Was the global default already added?

This one was 2 hours off:

![Image](https://github.com/ugent-library/biblio-backoffice/assets/1572885/fedcd294-ef00-44e2-8a4d-f15c2f4e6b35)
[...]

PR #1394, opened 2024-03-14
Feature: informative timestamps #1370

verheyenkoen:
This PR makes created/edited timestamps more informative as described in #1370.

### DIFF
--- a/views/shared/_created_by.gohtml
+++ b/views/shared/_created_by.gohtml
@@ -1,1 +1,1 @@
-Created {{formatTime .DateCreated nil "02/01/2006 15:04"}}{{if .Creator}} by {{.Creator.FullName}}{{ end }}.
+Created {{formatTime .DateCreated "02/01/2006 15:04"}}{{if .Creator}} by {{.Creator.FullName}}{{ end }}.
--- a/views/shared/_created_by.gohtml
+++ b/views/shared/_created_by.gohtml
@@ -1,5 +1,1 @@
-{{if .Creator}}
-Created {{timeElapsed .DateCreated "en"}} by {{.Creator.FullName}}.
-{{else}}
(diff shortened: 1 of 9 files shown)


views/shared/_created_by.gohtml
```diff
@@ -1,5 +1 @@
-{{if .Creator}}
-Created {{timeElapsed .DateCreated "en"}} by {{.Creator.FullName}}.
-{{else}}
-Created {{timeElapsed .DateCreated "en"}}.
-{{end}}
+Created {{formatTime .DateCreated nil "02/01/2006 15:04"}}{{if .Creator}} by {{.Creator.FullName}}{{ end }}.
```
verheyenkoen, line 1: @nics @netsensei @nicolasfranck Anyone knows how to get this working with the default timezone in devcontainers? Passing on `nil` for now as the `.TimeZone` in the dict coming from the `.Yield` method is returning `interface {}` (which is an invalid timezone).

Also tried setting `BIBLIO_BACKOFFICE_TIMEZONE="Europe/Brussels"` in ".devcontainer/.env" but that didn't seem to do anything (and shouldn't be necessary [if it's the default](https://github.com/ugent-library/biblio-backoffice?tab=readme-ov-file#configuration)).

nicolasfranck: `TZ=Europe/Brussels` should normally do something, and that is an environment variable that has an effect in any shell (set `export TZ=Europe/Brussels` and then do `date` in your shell). But normally the time zone should be set correctly in the destined docker environment. Any code that uses the "default time zone" should pick that up.

See also my comment here: https://github.com/ugent-library/biblio-backoffice/issues/935#issuecomment-1422568412

netsensei (approved):
Looks good to me.

PR #1398, opened 2024-03-19
Enhanced facet filter margins

verheyenkoen:
Enhanced facet filter margins with gap so there's no need to determine the last item and set margin dynamically.
CLOSES #1393

### DIFF
--- a/views/form/default/checkbox.gohtml
+++ b/views/form/default/checkbox.gohtml
@@ -1,6 +1,6 @@
 <div class="mb-6 row{{if .Error}} is-invalid{{end}}">
     <div class="col-{{.Cols}} offset-lg-3 mt-2">
-      <div class="form-check ">
+      <div class="form-check">
         <input class="form-check-input{{if .Error}} is-invalid{{end}}" type="checkbox" id="{{.Name}}" name="{{.Name}}"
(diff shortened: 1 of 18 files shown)


nicolasfranck (approved):

mietcls:
I gave feedback here @verheyenkoen: https://github.com/ugent-library/biblio-ugent-prototype/pull/17

verheyenkoen:
@mietcls Unless I'm missing something, the only difference I see between your change and what is in this PR is you removed a nested `<div>` which I tried at first but it doesn't work well (or needs to be fixed in some other way).

```html
<div class="dropdown-menu__body border-bottom d-flex flex-column gap-4">
    ...
</div>
```
instead of:
```html
<div class="dropdown-menu__body border-bottom">
    <div class="d-flex flex-column gap-4">
        ...
    </div>
</div>
```
This is what you get when you nest both divs (probably only a problem when facet texts are wrapped):
image

nics:
@mietcls @koen what's the status of this?

verheyenkoen:
Awaiting feedback from @mietcls

mietcls:
Correct @verheyenkoen – apologies. There's a non-urgent minor adjustment for the titles as well so they align with the select button, but I can add that optimisation later.

Issue #1402, opened 2024-03-25
[templ] Gohtml conversion to Templ


verheyenkoen:
This ticket is part of an effort to convert all Gohtml views to Templ. Some of these have already been converted but the bulk is yet to be addressed.

Most views are somehow dependent on other views which creates a virtual web of view dependencies which have to be converted all together which will be a massive effort and has to be concentrated on the shortest possible amount of time to avoid interference on other development. But some views (mostly loaded partially through HTMX, like eg. modal dialogs) can already be converted beforehand so that giant effort is reduced a little.

This ticket addresses those "smaller conversions" that don't impact the bulk of the views. We'll list the specific files that should be possible to convert individually (or with few related views - indented) here. We'll also try and identify unused views that can be removed here.

### Views to be converted

[✔️ = covered by Cypress tests] 

- [x] views/error_dialog.gohtml ✔️
  #1481 
- [ ] views/form_errors_dialog.gohtml
- [x] #1498
- [x] views/dataset/add_department.gohtml ✔️
  #1454
  - [x] views/dataset/_suggest_departments.gohtml ✔️
- [x] views/dataset/add_project.gohtml
  #1472 
  - [x] views/dataset/_suggest_projects.gohtml
- [x] views/dataset/add_publication.gohtml ✔️
  - [x] #1486
- [x] views/dataset/confirm_delete_abstract.gohtml ✔️
  #1444
- [x] views/dataset/confirm_delete_contributor.gohtml ✔️
  #1426 
_(other contributor related views have far more dependencies so are probably out of scope of this ticket)_
- [x] views/dataset/confirm_delete_department.gohtml ✔️
  #1461
- [x] views/dataset/confirm_delete_link.gohtml ✔️
  #1460 
- [x] views/dataset/confirm_delete_project.gohtml ✔️
  #1466 
- [x] views/dataset/confirm_delete_publication.gohtml ✔️
  #1465 
- [x] views/dataset/confirm_delete.gohtml ✔️
#1422 
- [x] views/dataset/confirm_publish.gohtml ✔️
  #1413 
- [x] views/dataset/confirm_republish.gohtml ✔️
  #1413 
- [x] views/dataset/confirm_withdraw.gohtml ✔️
  #1410 
- [x] views/media_types/suggest.gohtml ✔️
  #1424 
- [ ] views/modals/error.gohtml
- [ ] views/publication/add_dataset.gohtml ✔️
[...]

PR #1409, opened 2024-03-29
Base changes for refactor to Templ quick wins #1402

verheyenkoen:
Any generic changes that are required for other the conversion to Templ PR's go here.
REFS #1402


nics (approved):

views/views.go
```diff
@@ -0,0 +1,17 @@
+//go:generate go run github.com/a-h/templ/cmd/templ@v0.2.543 generate
+package views
+
+import (
+	"encoding/json"
+)
+
+func ToJSON(pairs ...string) string {
```
netsensei, line 8: @verheyenkoen So, this is only used for setting the snapshot ID as an `If-Match` header? e.g. ` hx-headers={ views.ToJSON("If-Match", dataset.SnapshotID) }`.  I don't think having a helper function that creates a simple key-value pair adds value here, unless we can argue that there's reuse elsewhere.

That's why the original code simply explicitly creates a JSON string: `hx-headers='{"If-Match": "{{.Dataset.SnapshotID}}"}'`. It's good enough to accomplish what it needs to do.

@nics

netsensei (changes requested):

views/views.go
verheyenkoen: Resolved using:
```go
hx-headers={ fmt.Sprintf(`{"If-Match": "%s"}`, dataset.SnapshotID) }
```

PR #1410, opened 2024-03-29
Refactored withdraw dataset confirm popup to templ #1402

verheyenkoen:
Depends on #1409


routes/register.go
```diff
@@ -292,6 +292,10 @@ func Register(c Config) {
 					// curator actions
 					r.Group(func(r *ich.Mux) {
 						r.Use(ctx.RequireCurator)
+
+						// withdraw
+						r.Get("/publish/withdraw", datasetEditingHandler.ConfirmWithdraw).Name("dataset_confirm_withdraw")
```
verheyenkoen, line 297: I've converted this as is, but shouldn't the route be `/withdraw/confirm` here?

nics (approved):

PR #1419, opened 2024-04-10
Refactored change publication type confirm popup to templ #1402

verheyenkoen:
Depends on #1409


routes/register.go
```diff
@@ -319,6 +319,10 @@ func Register(c Config) {
 				r.Route("/publication/{id}", func(r *ich.Mux) {
 					r.Use(ctx.SetPublication(c.Services.Repo))
 
+					// edit publication type
+					r.Get("/type/confirm", publicationEditingHandler.ConfirmUpdateType).Name("publication_confirm_update_type")
+					r.Put("/type", publicationEditingHandler.Wrap(publicationEditingHandler.UpdateType)).Name("publication_update_type")
```
nics, line 324: better to leave the ones that still need Wrap in their old router group, or 2 contexts will be set

nics (changes requested):

routes/register.go
verheyenkoen: @nics But the also need both contexts: https://github.com/ugent-library/biblio-backoffice/blob/687393107e69799adb28eec3bb409b1068528fc8/handlers/publicationediting/type.go#L34

verheyenkoen: @nics Reverted the `PUT /type` handler to the old model

netsensei (approved):

nics (approved):

PR #1431, opened 2024-04-18
clear session when user no longer found or inactive

nicolasfranck:
Use case: 

when the user record of the logged in user has been removed (which will probably never happen) or has been set inactive (more probable), the website crashes with an internal server error "could not get user from session", and no way to see what happened. Now this can only be fixed by clearing your cookies. I made sure that this clearing happens automatically.

Notes:

* user cache holds user record a little longer so in order to test you might want to restart the app after setting active:0 in the database.


nics:
@nicolasfranck as discussed, enhance HandleError and the ErrorHandlers registry  to handle specific errors like UserNotFound

nics (approved):

PR #1444, opened 2024-04-22
Refactored delete dataset abstract confirm popup to templ #1402

verheyenkoen:
blocked by https://github.com/ugent-library/biblio-backoffice/pull/1448


handlers/datasetediting/abstract.go
```diff
@@ -195,26 +192,25 @@ func (h *Handler) UpdateAbstract(w http.ResponseWriter, r *http.Request, ctx Con
...
-func (h *Handler) ConfirmDeleteAbstract(w http.ResponseWriter, r *http.Request, ctx Context) {
+func ConfirmDeleteAbstract(w http.ResponseWriter, r *http.Request) {
+	c := ctx.Get(r)
+	dataset := ctx.GetDataset(r)
+
 	var b BindDeleteAbstract
 	if err := bind.Request(r, &b); err != nil {
-		h.Logger.Warnw("confirm delete dataset: could not bind request arguments", "errors", err, "request", r, "user", ctx.User.ID)
+		c.Log.Warnw("confirm delete dataset: could not bind request arguments", "errors", err, "request", r, "user", c.User.ID)
 		render.BadRequest(w, r, err)
```
nics, line 202: should use c.HandleError with error httperror.BadRequest

nics (changes requested):
the temple command doesn't support subdirectories at the moment, so we need to keep all the files in the same directory for now (https://templ.guide/commands-and-tools/cli)

handlers/datasetediting/abstract.go
verheyenkoen: Done

Issue #1404, opened 2024-03-26
Align confirm delete modal popup header texts


verheyenkoen:
Change the card title for all confirm delete modal popups from "Are you sure" (may include question mark but not consistent) to "Confirm deletion".

#### Before
<img width="260" alt="image" src="https://github.com/ugent-library/biblio-backoffice/assets/599538/3bdd92ea-b1c9-44d3-b48a-767a9036784e">

#### After
![image](https://github.com/ugent-library/biblio-backoffice/assets/599538/dabc098b-1469-4df8-b8e4-ae3d68699493)

verheyenkoen:
We will group this with an effort to make a generic confirm delete dialog that can be used in many use cases. That can eliminate quite a bunch of custom confirm delete views.

verheyenkoen:
Yes, combined it with #1402 but all confirm delete dialogs are now converted to templ since today and also merged in dev.

PR #1449, opened 2024-04-23
Added a generic confirm delete dialog #1404

verheyenkoen:
This dialog will replace all current custom confirm dialog views (both gohtml and templ).

This PR will also align the delete modal dialog titles as described in #1404.


verheyenkoen:
No additional tests written. Everything here is already covered by the test for issue #1402.

nics (approved):

PR #1470, opened 2024-04-26
Fix/visual fixes

mietcls:
- Fixes missing padding since migration
- Adds icon to batch operations in sidebar


verheyenkoen (approved):
Looks good to me.

PR #1475, opened 2024-04-30
Refactor to templ/datasets

### LINKED ISSUE #1457: `views/search/*_facet.gohtml` en `views/search/_facets.gohtml`

nicolasfranck:
fixes #1457

Done:

* added templ views for dataset search
* added facet builders (to be used later for publications?)
* added `CurrentURL` to ctx.Ctx
* removed base handler from datasearching
* moved datasearching methods to new routing
* changed signatures views.PaginationCount and views.Pagination (it now accepts underlying pagination.Pagination instead of models.SearchHits to make it more broadly useable). View `views/candidaterecords/list.templ` was changed in the process
* added url (string) argument to contributorviews.Summary to link to tab where you can add/edit contributors (when not empty). View `views/publication/summary.templ` was changed in the process
* copied urls.linkify to views package


views/dataset/summary.templ
```diff
@@ -0,0 +1,204 @@
...
+}
+ 
+func getFirstDatasetIdentifier(dataset *models.Dataset) (string, []string) {
+    for typ, vals := range dataset.Identifiers {
+        return typ, vals
+    }
+    return "", nil
+}
+
+templ Summary(c *ctx.Ctx, args SummaryArgs) {
```
nics, line 29: @verheyenkoen @nicolasfranck just a general remark: we'll have to decide on an args style
- put everything  in an Args struct
- put everything except the ctx in an Args struct
- put everything except ctx and required arguments in an Opts struct
What are your thoughts?
Not a blocker, these things are easily refactored.

verheyenkoen: Since an args struct is a kinda-sorta solution to optional arguments I was thinking the last one but you'd still need to pass on nil for the Opts struct parameter if you don't want to pass any. I'd leave context separate anyway as the first argument.

nicolasfranck: * ctx.Ctx should always be the first argument. Even if we could embed it into a required Args. A bit like the context.Context that is used everywhere as the first argument. Lets not forget that for simpler cases with not much arguments an Args struct might be overkill, and in those cases we still have to supply the ctx. That way it is the same for the style with an Args struct. (`Method(c ctx.Ctx, args Args)`) or with separate arguments (`Method(c ctx.Ctx, arg1, arg2 ..)`)
* I'm not convinced that optional arguments should be all be in an optional variadic arguments like those Opts. These are arguments to templates, struct attributes are never required by the language, and what if there more than one optional types of arguments?

nics:
@nicolasfranck this needs some updates to use @verheyenkoen's url builder

views/dataset/summary.templ
nics: @nicolasfranck not variadic, just a struct as last argument
ex. `ShowPublication(ctx, pub, optionalStuff{})`

nics: but we have consensus on leaving ctx out of the args/opts

nics (approved):

views/record_timestamp.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/ugent-library/biblio-backoffice/models"
+)
+
+func CreatedBy(c *ctx.Ctx, createdAt *time.Time, createdBy *models.Person) string {
+	s := fmt.Sprintf("Created %s", createdAt.In(c.Timezone).Format("2006-01-02 15:04"))
+	if createdBy != nil {
+		s += " by " + createdBy.FullName
+	}
+	s += "."
+	return s
```
verheyenkoen, line 17: @nicolasfranck Maybe you can use [`lo.Ternary`](https://pkg.go.dev/github.com/samber/lo#readme-ternary) here?

PR #1476, opened 2024-05-01
Added URLBuilder for constructing outgoing url's

verheyenkoen:
Roughly related to #1402 

Currently only implemented these methods:

- Adapting methods:
  - `.WithPath(path ...string)`
  - `.AddQuery(key string, value string)`
  - `.WithQuery(query any)` ([see review](https://github.com/ugent-library/biblio-backoffice/pull/1476#discussion_r1585938615))
  - `.ClearQuery()`
- Build methods: 
  - `.URL() *url.URL`
  - `.String() string`
  - `.SafeURL() templ.SafeURL`

We can add more when we have use cases for them.

### DIFF
--- a/views/pagination.templ
+++ b/views/pagination.templ
@@ -9,7 +9,7 @@ templ Pagination(c *ctx.Ctx, baseURL *url.URL, searchArgs *models.SearchArgs, se
 	<ul class="pagination">
 		if searchHits.HasPreviousPage() {
 			<li class="page-item">
-				<a class="page-link" href={ URL(baseURL).WithQuery(searchArgs.Clone().WithPage(searchHits.PreviousPage())).SafeURL() } aria-label="Previous">
+				<a class="page-link" href={ URL(baseURL).Query(searchArgs.Clone().WithPage(searchHits.PreviousPage())).SafeURL() } aria-label="Previous">
 					<i class="if if-chevron-left" aria-hidden="true"></i>
 				</a>
 			</li>
@@ -23,7 +23,7 @@ templ Pagination(c *ctx.Ctx, baseURL *url.URL, searchArgs *models.SearchArgs, se
 		for _, page := range searchHits.PagesWithEllipsis() {
 			if page > 0 {
 				<li class={ "page-item", templ.KV("active", searchHits.Page() == page) }>
-					<a class="page-link" href={ URL(baseURL).WithQuery(searchArgs.Clone().WithPage(page)).SafeURL() } aria-label={ fmt.Sprintf("Page %d", page) }>
+					<a class="page-link" href={ URL(baseURL).Query(searchArgs.Clone().WithPage(page)).SafeURL() } aria-label={ fmt.Sprintf("Page %d", page) }>
 						{ fmt.Sprintf("%d", page) }
 					</a>
 				</li>
@@ -37,7 +37,7 @@ templ Pagination(c *ctx.Ctx, baseURL *url.URL, searchArgs *models.SearchArgs, se
 		}
 		if searchHits.HasNextPage() {
 			<li class="page-item">
-				<a class="page-link" href={ URL(baseURL).WithQuery(searchArgs.Clone().WithPage(searchHits.NextPage())).SafeURL() } aria-label="Next">
+				<a class="page-link" href={ URL(baseURL).Query(searchArgs.Clone().WithPage(searchHits.NextPage())).SafeURL() } aria-label="Next">
 					<i class="if if-chevron-right" aria-hidden="true"></i>
 				</a>
 			</li>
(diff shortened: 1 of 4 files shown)


views/pagination.templ
```diff
@@ -6,49 +6,49 @@ import "github.com/ugent-library/biblio-backoffice/ctx"
...
-            <a class="page-link" href="#" aria-label="Next">
-                <i class="if if-chevron-right" aria-hidden="true"></i>
-            </a>
-        </li>
-    }
-    </ul>    
+	<ul class="pagination">
+		if searchHits.HasPreviousPage() {
+			<li class="page-item">
+				<a class="page-link" href={ NewURLBuilder(baseURL.String()).WithQuery(searchArgs.Clone().WithPage(searchHits.PreviousPage())).SafeURL() } aria-label="Previous">
```
verheyenkoen, line 12: Already implemented for the pagination template but this is only used for candidate records which I don't know how to test. Can you test that, @nicolasfranck?

views/urls.go
```diff
@@ -3,15 +3,63 @@ package views
...
+type URLBuilder struct {
+	url *url.URL
+}
+
+func NewURLBuilder(base string) *URLBuilder {
+	url, err := url.Parse(base)
+	if err != nil {
+		// TODO: how to deal with this
+		panic(err)
+	}
```
verheyenkoen, line 19: Not sure how to deal with this. This intended to be used in views, so can't really return an error as second return value. But apparently `url.Parse` is very forgiving on what you provide as the base url so probably won't happen any time soon.

views/urls_test.go
```diff
@@ -0,0 +1,79 @@
+package views
```
verheyenkoen, line 1: Added a go test file for this. Is this intended to be run manually only or are these run in a pipeline as well?

views/urls.go
```diff
@@ -3,15 +3,63 @@ package views
...
-		return u
+		return builder
 	}
-	newU := *u
-	newU.RawQuery = vals.Encode()
-	return &newU
+
+	builder.url.RawQuery = vals.Encode()
+	return builder
+}
```
verheyenkoen, line 47: This is intended for use with a bindable struct (using `` `query:"param"` `` descriptors). Not sure if we should validate that the argument is indeed a bindable struct. If you pass a string eg, that would result in a weird query string:

```go
NewURLBuilder("https://example.com").WithQuery("foo=bar").String()

// becomes: https://example.com?=foo%3Dbar
```

views/urls.go
nics: 2 ways to think about this:
- we consider the wrong input a programming error, then panic is fine
- we consider the url as user input, then we need to handle the error

I think panic fine here

if you would like to handle the error:
templ, just like html/template allows a function to return `(string, error)`
so you could keep the error in an error struct field, and return at the end of the builder chain in `URL()`, `String()` and `SafeURL()`

views/urls.go
```diff
@@ -3,15 +3,63 @@ package views
...
 	"github.com/ugent-library/bind"
 )
 
-func urlWithQuery(u *url.URL, v any) *url.URL {
-	vals, err := bind.EncodeQuery(v)
+type URLBuilder struct {
+	url *url.URL
+}
+
+func NewURLBuilder(base string) *URLBuilder {
```
nics, line 14: we're going to use this a lot so maybe add a less verbose `function URL() *URLBuilder` constructor alias? And maybe also `Path` instead of `WithPath` etc?

views/pagination.templ
```diff
@@ -6,49 +6,49 @@ import "github.com/ugent-library/biblio-backoffice/ctx"
...
+					<i class="if if-chevron-left" aria-hidden="true"></i>
```
nicolasfranck, line 13: Is there a reason for these changing tab lengths? I use vscode too

verheyenkoen: That's what `templ fmt` did. Is that not right?

views/urls.go
verheyenkoen: @nics I just considered another option. How about accepting a `*url.URL` as base parameter here instead of `string`? Then handling the `url.Parse` error is out of scope of the builder and in most cases we'd be starting from a `*url.URL` anyway. Only in case of doing something like `NewURLBuilder(c.FrontendURL)...` you start from a string.

nics: good idea, or just have both

verheyenkoen: Done

views/urls.go
verheyenkoen: Done. Also renamed WithQuery to Query and AddQuery to AddQueryParam.

views/urls.go
nics: @koen Add a separate method `QueryString` or `RawQuery`?

verheyenkoen: Done

views/urls_test.go
nics: manual for now, but would like to change that

nics (approved):

nics (approved):

Issue #1411, opened 2024-04-02
[Plato imports] As a librarian, I can preview to-be-claimed dissertations


mietcls:
# Feature

As a librarian, I can preview to-be-claimed dissertations / suggestions

## Success criteria

- [ ] I can click "Preview" to see whether I want to import the dissertation or not
Click on preview: https://booktower-ugent-ui.vercel.app/biblio/reviewer/reviewer-suggestions.html
<img width="1390" alt="Screenshot 2024-04-26 at 10 57 40" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/568a00d1-fb81-4766-a6d8-a1d4cd8a3c4a">

- [ ] I can see the read-only version of the record in a modal 
  - that previews the record, without importing it
  - with a topbar that shows a citation-style title
  - with a clickable table of contents
- [ ] I can import the dissertation as a draft
<img width="1503" alt="Screenshot 2024-04-26 at 10 57 48" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/9a3005b1-3a0e-4ec7-aa62-29f9193c3cca">

PR #1482, opened 2024-05-03
[Plato imports] As a librarian, I can preview to-be-claimed dissertations #1411 

netsensei:
* Link to preview.
* Preview modal with dynamic fields implemented.
* Adds a dedicated file download handler for candidate records.


views/publication/summary.templ
```diff
@@ -36,7 +36,7 @@ templ Summary(c *ctx.Ctx, p *models.Publication, opts SummaryOpts) {
 									}
 								</span>
 								if f := p.MainFile(); f != nil {
-									@AccessLevel(c, f)
+									@SummaryAccessLevel(c, f)
```
nicolasfranck, line 39: I see that the candidaterecord publication preview here too starts too deviate too much from the publication summary in other places (which is more broadly used, but is older). I have rewritten that publication Summary for the older layout, and moved the newer publication preview into the candidaterecord view package.

See https://github.com/ugent-library/biblio-backoffice/pull/1484

PR #1499, opened 2024-05-07
Refactor to templ/dataset show activity

### LINKED ISSUE #1498: views/dataset/show_activity.gohtml

nicolasfranck:
fixes #1498

P.S. Is it necessary to set the main navigation for detail-pages) (dataset in this case? Seems to have no effect on those pages.
e.g. I have set the main navigation via middleware `ctx.SetNav` to "datasets", but if I change it to something
else (that does not exist), the menu "datasets" is still selected ;-)


routes/register.go
```diff
@@ -395,6 +395,10 @@ func Register(c Config) {
 					r.Use(ctx.RequireViewDataset)
 
 					// view only functions
+					// tab publications
+					r.With(ctx.SetNav("datasets"), ctx.SetSubNav("publications")).Get("/publications", datasetviewing.ShowPublications).Name("dataset_publications")
```
nics, line 399: you're setting Nav twice, it's already set higher up in the router hierarchy

nics (changes requested):

handlers/datasetviewing/show.go
```diff
@@ -90,10 +84,6 @@ func ShowPublications(w http.ResponseWriter, r *http.Request) {
...
 }
 
-func (h *Handler) ShowActivity(w http.ResponseWriter, r *http.Request, legacyCtx Context) {
-	render.View(w, "dataset/show_activity", YieldShowActivity{
-		Context:      legacyCtx,
-		SubNavs:      subNavs,
-		ActiveSubNav: "activity",
-	})
+func ShowActivity(w http.ResponseWriter, r *http.Request) {
+	datasetviews.ShowActivity(ctx.Get(r), ctx.GetDataset(r)).Render(r.Context(), w)
```
nicolasfranck, line 88: Keep the name `ShowActivity` or set it to a similar name (just `Activity`) like in the publicationviews?

nics: just Activity

routes/register.go
nicolasfranck: done

handlers/datasetviewing/show.go
nicolasfranck: done

nics (approved):

PR #1514, opened 2024-05-08
fix issue #1511

### LINKED ISSUE #1511: Refactor to templ: datasetviewing.ShowDescription
Enhances `datasetviewing.ShowDescription`

nicolasfranck:
fixes #1511


routes/register.go
```diff
@@ -399,6 +399,7 @@ func Register(c Config) {
 					r.Use(ctx.RequireViewDataset)
 
 					// view only functions
+					r.With(ctx.SetNav("datasets"), ctx.SetSubNav("description")).Get("/description", datasetviewing.ShowDescription).Name("dataset_description")
```
netsensei, line 402: Same remark as @nics. It's worth considering setting `ctx.SetNav` here at the group level, for instance, like this:

```
r.Group(func(r *ich.Mux) {
    r.Use(ctx.SetNav("datasets"))
    ...
})
```

netsensei (changes requested):

routes/register.go
nics: yes this should be a group, not individually like here:

`r.With(ctx.SetNav("publications")).Get("/publication", publicationsearching.Search).Name("publications")`

nicolasfranck: See https://github.com/ugent-library/biblio-backoffice/pull/1519

nicolasfranck: done

nics (approved):

PR #1515, opened 2024-05-08
do not translate values of facet reviewer_tags

nicolasfranck:
What it fixes: in a previous commit I added localization prefixes `reviewer_tags` when configuring the facet `reviewer_tags`. So if you have tags like `mytag` they now appear as `reviewer_tags.mytag`. This bug only appears in the route `/dataset`.


netsensei:
Looks good to me!

PR #1539, opened 2024-05-16
rename mutations;add remove_project mutation;bundle mutations in one update

### LINKED ISSUE #1512: Add project.remove to batch operations
# Enhancement
Add `project.remove` to batch operations for librarians.
Now we only have `project.add`.

## Expected behaviour (if we decide to make it a feature)
When going to https://backoffice.biblio.ugent.be/publication/batch, I can use an operation to remove projects from a publication.

## Suggestion
`8701504,project.remove,1234`
_recordidentifier,project.remove,projectidentifier_

In line with options such as https://booktower.gitbook.io/product-docs/producten-en-diensten/biblio-academische-bibliografie-en-repository/handleidingen/batch-operaties#voorbeelden-batch-operaties

nics:
resolves #1512

* natural mutation names (avoid yodaspeak)
* add remove_project mutation
* lock and unlock are replaced by `set_locked,true|false` for consistency (you win @mietcls)
* mutations on the same record that follow each other in the batch interface will be bundled together

this will result in only one update for 1234:
```
 1234,add_keyword,foo
 1234,add_project,xyz
```
but this will result in two:
```
 1234,add_keyword,foo
 2345,add_keyword,foo
 1234,add_project,xyz
```

before this is put in production:
* update gitbook docs
* update biblio manual


mietcls:
When I test with a faulty operation on line 1, the error says it could not process the last line instead of the first one:
Screenshot 2024-05-17 at 09 10 24
Screenshot 2024-05-17 at 09 10 28

Error handling for locking does not return the "locked value must be 'true' or 'false'" error
Screenshot 2024-05-17 at 09 14 16
Screenshot 2024-05-17 at 09 14 27

Now and then I get this error after trying to process, not sure why:
Screenshot 2024-05-17 at 08 57 18

mietcls:
Whenever I add a faulty operation on line 1, it processes the error as if it was a problem on the last line:

"setClassification" instead of "set_classification" on line 1:
Screenshot 2024-05-17 at 09 10 24

Error on line 20:
Screenshot 2024-05-17 at 09 10 28

Double check, if you leave out the first line, everything processes:
Screenshot 2024-05-17 at 10 10 10

mietcls:
- Lock-unlock error now works!
- Invalid CSRF token error hasn't come back.
- Error handling is now richer, but the line is not correct yet:
Screenshot 2024-05-17 at 16 25 53

nics:
> * Lock-unlock error now works!
> * Invalid CSRF token error hasn't come back.
> * Error handling is now richer, but the line is not correct yet:
> 
> Screenshot 2024-05-17 at 16 25 53

@mietcls refactored some more, should work correctly now

views/replace.templ
```diff
@@ -4,6 +4,18 @@ import (
 	"fmt"
 )
 
+templ Replace(selector string, component templ.Component) {
```
nicolasfranck, line 7: Have you merged the dev branch where this is now also available?

nics: @nicolasfranck done

repositories/repo.go
```diff
@@ -289,11 +293,20 @@ func (s *Repo) MutatePublication(id string, u *models.Person, muts ...Mutation)
 	}
 
 	for _, mut := range muts {
-		mutator, ok := s.config.PublicationMutators[mut.Op]
+		mutator, ok := s.config.PublicationMutators[mut.Name]
 		if !ok {
-			return fmt.Errorf("unknown mutation '%s'", mut.Op)
+			msg := fmt.Sprintf("unknown mutation '%s'", mut.Name)
+			if mut.Line != 0 {
```
nicolasfranck, line 299: Why is the difference between a mutation with a line set and one without?

nics: Line is set if the source of the mutations is line-delimited like a file or a textarea to give better error messages

nicolasfranck: ah human input versus machine induced input (e.g. cronjob)

nics: yes

nicolasfranck (approved):

PR #1547, opened 2024-05-17
dataset view (show/edit) did not show identifier(type) when manually created

nicolasfranck:
In the current dev branch, if you create a dataset **manually**, the fields identifier_type and identifier are missing
in the description (both show/edit).

Reason for this is the if-statement to get identifier and identifier_type out of the map [Identifiers](https://github.com/ugent-library/biblio-backoffice/blob/dev/models/dataset.go#L54):

* https://github.com/ugent-library/biblio-backoffice/blob/dev/views/dataset/details.templ#L68
* https://github.com/ugent-library/biblio-backoffice/blob/dev/views/dataset/edit_details.templ#L67

(P.S. there is no other way to assign variables in templ)

I replaced the `identifierType != ""` by `true` which is always true.

Not sure if we should keep storing the identifier like this

@verheyenkoen detected this

### DIFF
--- a/cypress/e2e/issues/issue-1237.cy.ts
+++ b/cypress/e2e/issues/issue-1237.cy.ts
@@ -675,7 +675,7 @@ describe("Issue #1237: Accessibility and mark-up: make sure labels are clickable
         cy.visitDataset();
 
         cy.updateFields("Dataset details", () => {
-          cy.intercept("PUT", "/dataset/*/details/edit/refresh-form*").as(
+          cy.intercept("PUT", "/dataset/*/details/edit/refresh*").as(
             "refreshForm",
           );
           cy.setFieldByLabel("License", "The license is not listed here");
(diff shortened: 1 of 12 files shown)


verheyenkoen (approved):

nics:
@nicolasfranck no if's that are always true; just add a method to dataset or something

nicolasfranck:
@nics ok, I have done the following:

* dataset.Identifier is always required, even if status is not public
* add errors for both `identifier_type` and `identifier`, because one of both can be missing during edit, and we need to show an error for the field where the error is.
* in the form, use key and value "as is" instead of only using the value when the identifierType is in the valid list, as it was done in production. This caused the form to clear the filled in value for "identifier" when "identifier_type" was not a valid value, or was empty.
* generate display/fields in a golang function and then add the components in the form.

nics:
@mietcls @nicolasfranck why should identifier be required for a draft? this was a simple issue about 2 fields that didn't get rendered

mietcls:
It isn't required indeed for drafts, no need to change it - only on publish.

Sorry @nicolasfranck if I caused any confusion - for datasets it is mostly useless without an identifier but we don't need to stop people from making a draft without it.

nicolasfranck:
@nics updated for review

nics (approved):

PR #1567, opened 2024-05-21
Remove river and replace with (cron) cli commands

nics:


cli/server_cmd.go
```diff
@@ -79,22 +76,8 @@ var serverStartCmd = &cobra.Command{
 		// 	defer ffclient.Close()
 		// }
 
-		// start jobs
-		riverClient, err := jobs.Start(context.TODO(), jobs.JobsConfig{
-			PgxPool:       services.PgxPool,
```
netsensei, line 84: Remove `services.PgxPool` from `types.go` as well. I think it's only used for river.

netsensei (changes requested):

netsensei (approved):

PR #1569, opened 2024-05-22
Centralize error logging + Use chi httplog + Better errors

### LINKED ISSUE #1572: Centralize error logging and improve error messages
goals:

- ctx.HandleError can produce a meaningful logging error message so we can remove most of the log messages in the handlers
- wrap errors in repo, index, importer, ... packages to improve debugging en give HandleError enough info to make a meaningful message
- use slog.Logger and chi httplog everywhere instead of zap so we don't need top maintain our own http logging package anymore (https://github.com/ugent-library/zaphttp)

preparatory work (already done):

- extend ich router wrapper with a method to retrieve the current route name so we can include it in log messages
- extend httperror package with a Wrap sugar method so we can easily include the original error in a http error

see #1569 for wip

TODO:

- [x] #1581
- [x] wrap repo errors
- [x] #1575
- [x] #1574
- [x] #1573
- [x] #1580
- [x] #1576
- [x] #1585
- [x] #1584
- [x] #1579
- [x] #1586 (except for grpc)

nics:
resolves #1572

note: zap is still in use, but only for grpc logging


handlers/publicationediting/abstract.go
```diff
@@ -85,15 +83,14 @@ func EditAbstract(w http.ResponseWriter, r *http.Request) {
...
-		c.HandleError(w, r, httperror.BadRequest)
+		c.HandleError(w, r, httperror.BadRequest.Wrap(err))
 		return
 	}
 
 	abstract := p.GetAbstract(b.AbstractID)
 
 	if abstract == nil {
-		c.Log.Warnf("edit publication abstract: Could not fetch the abstract:", "publication", p.ID, "abstract", b.AbstractID, "user", c.User.ID)
+		c.Log.Warn("edit publication abstract: Could not fetch the abstract:", "publication", p.ID, "abstract", b.AbstractID, "user", c.User.ID)
```
netsensei, line 93: Could => could (caps)

netsensei (changes requested):

backends/es6/publication_index.go
```diff
@@ -310,16 +310,16 @@ func (pi *PublicationIndex) DeleteAll() error {
...
 	defer res.Body.Close()
 
 	if res.IsError() {
 		buf := &bytes.Buffer{}
 		if _, err := io.Copy(buf, res.Body); err != nil {
-			return err
+			return fmt.Errorf("publicationindex.DeleteAll: io error while reading es6 error response body: %w", err)
 		}
-		return errors.New("Es6 error response: " + buf.String())
+		return errors.New("publicationindex.DeleteAll: es6 error response: " + buf.String())
```
verheyenkoen, line 322: Out of curiosity: why uses `errors.New` here and not `ftm.Errorf`?

nicolasfranck: If it is just to concatenate strings, I do not see any gain in using `fmt.Errorf`. 

`Errors.New("my " + "concatenated string")` 

vs

`fmt.Errorf("my %s", "concatenated string")`

If you trying to insert anything other than strings (float, integer, and even a wrapped error), that is another case. `Errors.New("my length: " + fmt.Sprintf("%d", 186))` would make it too verbose, so `fmt.Errorf("my length: %d", 186)` is more concise.

verheyenkoen (approved):

netsensei (approved):
Looks good to me.

PR #1592, opened 2024-06-03
Asset building: replace laravel-mix with esbuild

verheyenkoen:


esbuild.mjs
```diff
@@ -0,0 +1,69 @@
...
+import { sassPlugin } from "esbuild-sass-plugin";
+import manifestPlugin from "esbuild-plugin-manifest";
+import fs from "fs";
+
+const config = {
+  entryPoints: [
+    { in: "assets/js/app.js", out: "js/app" },
+    { in: "assets/css/app.scss", out: "css/app" },
+    // Entry points outside the "ugent" folder should be mapped manually to avoid transferring the "ugent" folder to the output directory
+    { in: "assets/images/plato-logo.svg", out: "images/plato-logo.svg" },
```
nics, line 12: @verheyenkoen can't this contain a wildcard like `{ in: "assets/images/*svg", out: "images" }`?
or we can move everything to a subfolder so we don't have to list very image?

verheyenkoen: @nics Nope spent 3 hours trying to fix that. If we would add a second image to that folder, I think wildcards are gonna work but if it's just one file (even with wildcard), the `out` is assumed to be the file name used, not the folder name.  Out also doesn't support backreference substitution of any kind.

There were 2 other possible solutions but not sure if we can do those:
* Move the file to `assets/ugent/images` with the others, so the "ugent" level can be asummed the root level for assets that are not bundled into one file (everything that is not JS/CSS).
* Allowing the "ugent" folder level to be copied to the static folder but then we'd also need a fix for the manifest keys.

nics: or just move the images in ugent/images to a general images folder with everything?

verheyenkoen: Ok, I'll do that.

nics (approved):

PR #1595, opened 2024-06-07
Prepare dev for release

nics:
- people and projects work saved in the https://github.com/ugent-library/biblio-backoffice/tree/people_projects branch
- revert to authority client
- update dev container for seamless experience


netsensei (approved):
Tested this and it works for me.

backends/authority/organization.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		if res.IsError() {
+			return fmt.Errorf("%+v", res)
+		}
+		time.Sleep(5 * time.Second)
+	}
+	return nil
+}
+
+func (c *Client) SeedOrganization(data []byte) error {
```
nicolasfranck, line 61: Why do you provide a raw byte array instead of an Organization

backends/authority/organization.go
```diff
@@ -0,0 +1,166 @@
...
+		return fmt.Errorf("%+v", res)
+	}
+	return nil
+}
+
+func (c *Client) CountOrganizations() (int64, error) {
+	return c.mongo.Database("authority").Collection("organization").CountDocuments(context.Background(), bson.D{}, options.Count().SetHint("_id_"))
+}
+
+func (c *Client) GetOrganization(id string) (*models.Organization, error) {
```
nicolasfranck, line 90: Better not retrieve the organization record from mongodb directly, and leave access via elasticsearch for more complex search actions? GetPerson does that too for example

backends/authority/person.go
```diff
@@ -0,0 +1,245 @@
...
+			return err
+		}
+		if res.IsError() {
+			return fmt.Errorf("%+v", res)
+		}
+	}
+	return nil
+}
+
+func (c *Client) SeedPerson(data []byte) error {
```
nicolasfranck, line 45: Why do you provide a byte array instead of a Person object?

backends/authority/project.go
```diff
@@ -0,0 +1,262 @@
...
+			return err
+		}
+		if res.IsError() {
+			return fmt.Errorf("%+v", res)
+		}
+	}
+	return nil
+}
+
+func (c *Client) SeedProject(data []byte) error {
```
nicolasfranck, line 41: Why provide a byte array instead of a model Project?

backends/authority/organization.go
nics: there are no organizations in the mongo db

backends/authority/person.go
nics: they are not serializations Person objects, but the structure that the frontoffice expects

backends/authority/organization.go
nics: see comment below

nicolasfranck (approved):

Issue #1278, opened 2023-10-19
[Plato imports] As a researcher or supervisor, I can see how I am related to a record (supervisor, author, editor, ...) in the card


mietcls:
# Enhancement

Part of Plato imports, but is a feature in itself.

As a researcher or supervisor, I can see how I am related to a record (supervisor, author, editor, ...) in the card. 
_The distinction is needed, because supervising is not the same as contributing as an author/editor, but important information nonetheless._

- [x] For each record, I can see which role I had to play:
  - Publications
    - [x] `Your role: author`
    - [x] `Your role: supervisor`
  - Datasets
    - [x] `Your role: creator`
  - No official role in the record:
    - [x] Registrar (Librarian and Non-librarian): `Your role: registrar`
    - [x] Librarian reviewing / curating: _no role_
- [x] In the search and suggestions overview; e.g.:
`<div class="c-author"><span class="badge rounded-pill badge-light">Your role: supervisor</span></div>`

Publication / dataset record
<img width="1341" alt="Screenshot 2024-06-11 at 16 19 29" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/94080e51-d8e0-4947-b357-1074e6640b35">

Suggestion (currently only deployed for reviewers, so it will be less relevant – in the future it will be for researchers)
![Screenshot 2024-06-11 at 16 03 44](https://github.com/ugent-library/biblio-backoffice/assets/1572885/7d5ecb6b-6c19-416a-8324-aed464055f75)

## Context
Decided to not put it in the header of the edit / view mode; as this would clutter the header again.

## Out of scope

- `Your role: editor`

PR #1599, opened 2024-06-11
[Plato imports] Contributor role labels #1278

verheyenkoen:
REFS #1278


models/dataset.go
```diff
@@ -173,6 +172,18 @@ func (d *Dataset) RemoveContributor(role string, i int) error {
 	return nil
 }
 
+func (d *Dataset) GetUserContributorRoles(user *Person) string {
```
nics, line 175: this shouldn't be a method on dataset, it belongs in a view helper or in the io18n dictionary

models/person.go
```diff
@@ -217,3 +220,9 @@ func (u *Person) CanViewDashboard() bool {
 func (u *Person) CanChangeType(p *Publication) bool {
 	return u.CanEditPublication(p) && p.Status != "public"
 }
+
+func (p *Person) IsContributor(c []*Contributor) bool {
```
nics, line 224: likewise, this shouldn't be a method, a plain function will do

if it needs to be method, dataset/publication.HasContributor(p) makes more sense

models/publication.go
```diff
@@ -431,6 +430,28 @@ func (p *Publication) RemoveContributor(role string, i int) error {
 	return nil
 }
 
+func (p *Publication) GetUserContributorRoles(user *Person) string {
```
nics, line 433: see remark above, it's a view helper

nics (changes requested):

views/dataset/summary/summary.templ
```diff
@@ -5,25 +5,43 @@ import (
...
+	Target  string
 	Actions templ.Component
 	Footer  templ.Component
 	Links   templ.Component
 }
 
+func getUserContributorRoles(d *models.Dataset, user *models.Person) string {
+	if d.HasContributor("author", user) {
+		return "creator"
+	}
```
verheyenkoen, line 25: This is a bit tricky though. If we ever rename dataset authors to creators (as is marked with a TODO), this could easily be forgotten (but a test would fail).

netsensei: @verheyenkoen It does make sense to add it as a separate view helper function because you'd end up with a complex block of conditionals in template otherwise. However, I think this function is a smell of two deeper issues we won't change on short notice.

1/ The `author` role to `creator` transformation is a band-aid to avoid a data transformation / code refactor. Regardless of where we put it, it's something to keep in mind later on anyway.

2/ I'd try making the case for model methods as a sort of "calculated / dynamic" properties for myself, but in our architectural design, I don't think that belongs at the model level such as it currently is. The `registrar` check drives that point  home: it's an example of why directly exposing an object from the data layer to the view layer creates inflexible tight coupling. Suppose we were to create a separate "view object" (or DTO) for a summary containing only relevant data to that specific view - instead of passing the entire publication / dataset - then we could add specific getter methods on that object and treat them as dynamic properties in the templ views. However, that's a significant change to our architecture, and something we need to think about first.

cc @nics

nics: this is all just about labels (a dataset author has as label "creator"), so it belongs in views and i18n

nics: @verheyenkoen just can remove the TODO about the field name, it won't happen and is not important

netsensei: @nics I'm not convinced it's just a label. It's not important right now, so let's discuss this later on.

nics (approved):

Issue #1632, opened 2024-06-25
Message URL not rendered correctly on Biblio messages tab


mietcls:
# Bug

Message URL not rendered correctly on Biblio messages tab. It does look correct on edit and on the search overview page.

Context:
<img width="1413" alt="Screenshot 2024-06-25 at 16 31 39" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/32f27064-8d75-4ddb-a26c-5bd1f11972fc">

Bug:
<img width="679" alt="Screenshot 2024-06-25 at 16 31 36" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/3dddc3bf-be22-4f58-9d53-e194670bf190">

PR #1635, opened 2024-06-26
Fix biblio message links not showing for publications in activity tab

mietcls:
Fix issue #1632 and add changelog starting from June

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Removed Laravel Mix and added [esbuild](https://esbuild.github.io/) for asset bundling
+- Add [Plato import view for librarians](https://github.com/ugent-library/biblio-backoffice/issues/1231)
+- Replaced templating with [Templ](https://templ.guide)
+- Align how you can add tags and keywords to records, by using the [Tagify](https://github.com/yairEO/tagify) plugin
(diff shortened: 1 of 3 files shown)


verheyenkoen (approved):

verheyenkoen:
@mietcls This would be a good case for writing a first test (also for datasets and also in the overview). You can add these to `cypress/e2e/datasets/edit/biblio-messages.cy.ts` resp. `cypress/e2e/publicationss/edit/biblio-messages.cy.ts`. Perhaps first do this on a separate branch so this branch isn't blocked from merging.

Hints: 
- Use the `cy.updateFields` command [like here](https://github.com/ugent-library/biblio-backoffice/blob/7111e67546c65c6876df0a59145b9af5c9d9d158/cypress/e2e/datasets/edit/biblio-messages.cy.ts#L43). Remember this command takes a third parameter to indicate wether changes should be persisted which is false by default. Pass `true` to have changes persisted (or the save button text if it is not just "Save").
- Use the `cy.setFieldByLabel` command [like here](https://github.com/ugent-library/biblio-backoffice/blob/7111e67546c65c6876df0a59145b9af5c9d9d158/cypress/e2e/datasets/edit/biblio-messages.cy.ts#L21).
- Use the [`cy.get`](https://docs.cypress.io/api/commands/get) and [`cy.find`](https://docs.cypress.io/api/commands/find) commands and `.should("have.attr", "...", "...")` assertion to verify that a hyperlink is rendered where you expect it.

PR #1636, opened 2024-06-26
Fix departments not showing and extra metadata line showing when needed

mietcls:
- Fixes departments showing for publications
- Shows VABB, Legacy and linked datasets/publications when one of those metadata points is available, and hides the metadata line if none of that data is available.
- Aligns UI behaviour for publications and datasets


views/publication/summary/summary.templ
```diff
@@ -162,26 +162,28 @@ templ Summary(c *ctx.Ctx, args SummaryArgs) {
...
+											target={ args.Target }
+										}
+									>
+										<i class="if if-edit if--small"></i>
+										<em>Add department</em>
+									</a>
+								}
+							</div>
+						</div>
+						if args.Publication.VABB() != "" || args.Publication.Legacy || len(args.Publication.RelatedDataset) > 0 {
```
mietcls, line 185: I can only add VABB year via batch (works as intended), I cannot fake information such as `VABB: c:vabb:495300, VABB-1, approved, 2021`. Will this piece of code also show that information when available, and not just the year?

nics: yes args.Publication.VABB() will should all the info as a string

nics (approved):

PR #1658, opened 2024-07-11
Refactor reflex to wgo

verheyenkoen:
Removed reflex in favor of wgo (also with templ auto reload version)


Makefile
```diff
@@ -1,70 +1,37 @@
...
+	--proxy="http://localhost:3011" \
+	--proxyport=3001 \
+	--proxybind="localhost" \
+	--open-browser=false
+
+live/esbuild:
+	@node esbuild.mjs --watch
+
+live/server:
+	@BIBLIO_BACKOFFICE_PORT=3011 go run $(vWgo) run \
```
verheyenkoen, line 17: I chose to run the server (in live mode only) on port 3011 so the templ proxy can run on 3001 which aligns with dev mode.

INSTALL.md
```diff
@@ -259,30 +258,42 @@ npm install
...
+time=2024-07-11T11:57:05.993+02:00 level=INFO msg="starting server at localhost:3001"
 ```
 
 Open your browser and navigate to `http://localhost:3001`. You should see the application.
 
+> **IMPORTANT!**
+>
+> When running in live reload mode, Templ will generate different .go files along with a .txt file for every .templ file.
+> The .txt files are ignored via .gitignore but the .go files cannot be ignored. Make sure to never commit these
+> and shut down the live reload server before you commit any changes to .templ files.
```
verheyenkoen, line 292: In case this causes problems (someone commits the wrong .go files) we can adapt the [js-precommit-hook](https://github.com/ugent-library/js-precommit-hook) library to allow configurable substrings to block commit for (in this case the "WriteWatchModeString" substring which templ generates in live reload mode only).

Makefile
```diff
@@ -1,70 +1,37 @@
-export GO111MODULE=on
```
netsensei, line 1: Is there a particular reason to wholesale delete the scripts to initiate unit testing of the gRPC logic right now, via this PR which is focuses on live reloading?

netsensei (commented):
Tested this. Works fine for me. Only remark is related to the removal of the make scripts for unit testing of the gRPC endpoint.

Makefile
verheyenkoen: At first I didn't intend to do this, but then I got errors while running the scripts (I believe it was something related to the setting of the ENV vars) and since @nics already declared all that was not used anymore I deleted it.

nics: we'll move them to server/Makefile

Issue #1338, opened 2023-12-05
[quick import] Make publication detail header more informative


mietcls:
# Enhancement
Make publication detail header more informative: show metadata in the header MLA citation style.

## Expected behaviour

Show metadata in the header MLA-citation-style-inspired:
![Screenshot 2023-12-05 at 10 12 56](https://github.com/ugent-library/biblio-backoffice/assets/1572885/ba5759e0-1a1a-4f54-96f9-8b8bd4a04070)

**Author/editor information**
- [x] Journal article, Book chapter, Book, Conference contribution, Dissertation, Miscellaneous
  - _e.g. Last name, First name, et al._
  - `Name of first author` + `,`
  - `et al.` + `,` if multiple authors
- [x] Book editor / issue editor
  - _e.g. Last name, First name, et al., editors_
  - `Name of first editor` + `,`
  - `et al.` + `,` if multiple editors
  - `editor` if only one editor, `editors` if multiple
- [x] Dataset (e.g. Last name, First name, et al., creators)
  - _e.g. Last name, First name, et al., creators_
  - `Name of first creator` + `,`
  - `et al.` + `,` if multiple creators
  - `creator` if only one creator, `creators` if multiple

**Publications specific:**

- **Publication details** 
  - [x] Title
  - [x] Publication type
  - [x] Access level
  - [x] Publication year
  - [x] Acces level

- MLA style citation
  - [x] Publisher (for books / book chapters / book editors / dissertations / misc)
  - [x] Book (for book chapters)
  - [x] Conference (for conferences)
  - [x] Journal title (for journal articles / issue editor)
  - [x] Prodeedings title (for conferences)
  - [x] Publication title (for misc)
  - [x] Volume (for issue editor / misc)
  - [x] Pages
  - [x] Issue (for issue editor)

**Datasets specific:**

- **Datasets details** 
  - [x] Title
  - [x] Publisher
  - [x] Publication year
  - [x] Identifier type
  - [x] License
  - [x] Acces level

**Biblio information**
- [x] Status
- [x] Classification (publications only)
- [x] Biblio ID

**Extra**
- [x] Move buttons, remove "back to"
- [x] Fix this copy 
![Screenshot 2024-07-18 at 12 37 14](https://github.com/user-attachments/assets/29650d27-7dba-4192-bba7-2d1034d69bb9)
- [x] Align icons
[...]

PR #1659, opened 2024-07-12
Added dynamic breadcrumbs #1337

verheyenkoen:
Added breadcrumb functionality via PageLayout arguments, eg:

```go
templ Show(c *ctx.Ctx, p *models.Publication, redirectURL string) {
	@views.PageLayout(c, views.PageLayoutArgs{
		Title: c.Loc.Get("publication.page.show.title"),
		Breadcrumbs: []views.Breadcrumb{
			{LabelID: "publications", URL: c.PathTo("publications")}, // Breadcrumb with link
			{LabelID: "publication"}, // Breadcrumb without link
		},
	}) {
            // ...
```

Each breadcrumb should be translated in a .po file with the context "breadcrumbs" using this format:

```pot
msgctxt "breadcrumbs"
msgid "dashboard"
msgstr "Dashboard"
```

FIXES #1337

Issue #1623, opened 2024-06-19
Biblio issues after June release 


mietcls:
# Bug description
Biblio issues before June release

_[do not close, will keep editing]_

## Functional bugs

- [ ] Console error for candidate records?
Happens after publishing a record, afterwards the yellow dot recovers
<img width="752" alt="Screenshot 2024-06-17 at 12 08 23" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/5b3f071b-5bff-4acb-88cb-da4fb324f676">
<img width="880" alt="Screenshot 2024-06-17 at 11 51 10" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/552ede4c-bb48-4e4f-8d19-a688b2e10a4a">

## Visual bugs (Miet)
- [ ] Fix icons for file uploads  ![Screenshot 2024-06-13 at 17 02 28](https://github.com/ugent-library/biblio-backoffice/assets/1572885/abe700d4-ca81-4bb2-a539-11c55858562f)
- [ ] Fix double line when there is no content ![Screenshot 2024-06-13 at 17 04 29](https://github.com/ugent-library/biblio-backoffice/assets/1572885/e174876d-ea3b-41e0-828f-c2a59f9acb89)
- [ ] Missing blankstate <img width="212" alt="Screenshot 2024-06-14 at 15 45 07" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/6791140e-4876-41d4-bc0d-43815d56b9f5">
- [ ] Repeat locked message on tabs to make sure you know what is going on
- [ ] Reset filters: make sure tab does not reset, as it is not in scope
- [x] #1620
<img width="1212" alt="Screenshot 2024-06-17 at 09 38 30" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/8d9200d3-359c-4183-90fe-2795fa4c4c27">
<img width="722" alt="Screenshot 2024-06-17 at 10 08 41" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/203ae280-2d08-4885-8831-259a3f97359a">

- [ ] Make issue: add VABB to detail view reviewers 
<img width="1171" alt="Screenshot 2024-06-17 at 09 54 04" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/cf846667-d482-494e-9631-5a7d067c1921">

- [ ] Remove superfluous comment for books, book editor, issue editor and dissertations
[...]

PR #1661, opened 2024-07-16
Extend detail page summary

mietcls:
Depends on https://github.com/ugent-library/biblio-backoffice/pull/1659 – needs to merged first.

1. Made detail page summary in header more informative
Fixes https://github.com/ugent-library/biblio-backoffice/issues/1338

🙏 Not happened yet: live updating of header after editing. Could use some help with that, can be in a separate PR.

- publication: citation style header with
  - authors or editors
  - title
  - Journal title 
  - year
  - page count / volume / issue / first and last page
  - publisher
  - conference name if applicable
- datasetscitation style header with
  - contributors
  - title
  - publisher
  - year
  - identifier type
- Biblio ID
- Repositioning of header elements and responsiveness

2. Aligned access level icons and copy in search overview and full text & files view
Task part of https://github.com/ugent-library/biblio-backoffice/issues/1623

TODO: 
- dataset DOI double check
- double check names for different publication types


views/publication/show.templ
```diff
@@ -9,157 +9,237 @@ import (
...
+											<i class="if if-lock if--small if--muted"></i>
+											<span class="text-muted c-body-small ms-2">Locked</span>
+										</span>
+									}
+									<span class="c-subline text-nowrap me-3 pe-3 border-end">
+										{ c.Loc.Get("publication_types." + p.Type) }
+										if p.Classification != "" {
+											{ p.Classification }
+										}
+									</span>
```
verheyenkoen, line 30: Publication type and classification are concatenated with just a space this way which feels odd to mee. I see the prototype doesn't include the classification so not sure if this is ok, but wanted to note that it doesn't look right anyway:

image

views/publication/show.templ
```diff
@@ -9,157 +9,237 @@ import (
...
+									// TODO: figure out whether this is still needed
+									// if p.PublicationStatus != "" {
+									// 	<span class="c-subline text-nowrap me-3 pe-3 border-end">{ c.Loc.Get("publication_publishing_statuses." + p.PublicationStatus) }</span>
+									// }
```
verheyenkoen, line 34: Just wanted to tag this bit so we don't lose track and it would be removed before merge if not needed anymore.

views/publication/show.templ
```diff
@@ -9,157 +9,237 @@ import (
...
+											<div class="btn-text">Publish <span class="d-md-none d-lg-inline">to Biblio</span></div>
+										</button>
+									</div>
+								}
+							</div>
+						</div>
+						<h4 class="w-100 mb-0 mb-lg-2">
+
+							if (p.Type == "issue_editor") || (p.Type == "book_editor") {
+								if len(p.Editor) == 1 && p.Editor[0] != nil {
```
verheyenkoen, line 163: `nil` checks should not be necessary here (and on lines 167 and 171). They should never be `nil` as discussed [a while back](https://github.com/ugent-library/biblio-backoffice/commit/f068df91beec58fef14ca9d3812765ba5ec9f7ac#r144323945).

views/publication/show.templ
```diff
@@ -9,157 +9,237 @@ import (
...
+									{ p.Editor[0].LastName() }, { p.Editor[0].FirstName() }, editor.
+								}
+
+								if len(p.Editor) >= 2 && p.Editor[0] != nil {
```
verheyenkoen, line 167: I prefer `len(p.Editor) > 1` above `len(p.Editor) >= 2`*, but if you remove the `nil`-check this is gonna be an `else` anyway.

(*) Easier to comprehend and I mist the "=" which led me to believe it was a bug at first.

verheyenkoen (changes requested):

verheyenkoen (changes requested):
2 tests fail because of these changes (issue-1370).

views/publication/show.templ
mietcls: If we just `else` it woud also show it when there are 0 editors, which would cause it to break – but I'll replace the complex syntax!

mietcls:
> 2 tests fail because of these changes (issue-1370).

This one will change as well when the info sidebar is there, to save space and only focus on the important info: https://github.com/ugent-library/biblio-backoffice/issues/1663
We would only show `edited` in the header, and the other info would move into the info sidebar.

mietcls:
@verheyenkoen fixed with [6abd3e9](https://github.com/ugent-library/biblio-backoffice/pull/1661/commits/6abd3e90e6125461dbd2b2081b00f6721f0b9685)

verheyenkoen (approved):

nics (approved):

PR #1664, opened 2024-07-23
Hide curator names #1258

### LINKED ISSUE #1258: Hide Biblio team member names for researchers
# Enhancement
Hide Biblio team member names for researchers only

- [x] For researchers: Instead of the name, show "A Biblio team member"
- [x] For librarians: Librarians _should_ be able to see the names of each other.
- [x] In the publications and datasets search overview
Screenshot 2023-10-13 at 09 06 50
- [x] On record level for datasets and publications
Screenshot 2023-10-13 at 09 11 18
- [x] In the upcoming activity features; epic #1225 

## Context

When a librarian adapted a record, their names should not be open to researchers immediately to protect their privacy. This also makes sure researchers contact a generic helpdesk instead of people.

## Out of scope

- Inside the helpdesk it can be clear there is a person behind the screen, but they should not be able to "shoot" the messenger.
- Rejected name: "[A metadata librarian](https://journals.ala.org/index.php/lrts/article/view/4906/5926)" (too much of a library perspective, less product- and service-first)
- People that _used_ to be librarians

verheyenkoen:
Hide curator names in publication/dataset summary for non-curators.

Also includes some small changes in recent activity on the dashboard:
- aligned the existing behaviour (didn't show curator names to other curators)
- added unlock activity
- better lock icon
- removed duplicate activity entry if Biblio message was altered

FIXES #1258


mietcls (commented):
All changes behave as expected, thanks. However, I think I was wrong in our discussion before about this one:
> removed duplicate activity entry if Biblio message was altered

It now also deduplicates edits. If you edit, and then lock something the "edit" activity is hidden as well which is information people would want to have. Is this a big change?

views/record_timestamp.go
```diff
@@ -1,30 +1,56 @@
...
+	"strings"
 	"time"
 
 	"github.com/ugent-library/biblio-backoffice/ctx"
 	"github.com/ugent-library/biblio-backoffice/models"
 )
 
 func CreatedBy(c *ctx.Ctx, createdAt *time.Time, createdBy *models.Person) string {
-	s := fmt.Sprintf("Created %s", createdAt.In(c.Timezone).Format("2006-01-02 15:04"))
+	sb := strings.Builder{}
```
nics, line 12: better to store a pointer with `&strings.Builder{}`, all methods of builder are pointer methods: https://pkg.go.dev/strings#Builder.WriteString

nics (changes requested):

nicolasfranck:
An unrelated question @mietcls : shouldn't there be a bit more difference between the plain text "Edited by <user>" and the link "<title-with-link>" in the list of recent activity? Now it does not really look like there is something clickable.

views/record_timestamp.go
verheyenkoen: Fixed in 176b383ba71f58c9db7240d5e809c84700e6635d

nics (approved):

verheyenkoen:
> All changes behave as expected, thanks. However, I think I was wrong in our discussion before about this one:
> 
> > removed duplicate activity entry if Biblio message was altered
> 
> It now also deduplicates edits. If you edit, and then lock something the "edit" activity is hidden as well which is information people would want to have. Is this a big change?

The problem is that "edited" in the recent activity list is dubious. We determine "edited" as in "any change that we do not reduce as a special change (lock/publish/message/...)". With our current setup, if you change publication details and then set a message that is basically the same as changing the publication weeks ago and then setting a message as the next action now. The only difference is that the snapshot dates are further apart. But you can't really show the edited in the second scenario as there will be much more recent activity in other publications that is more recent and relevant.

The basic setup is wrong here where we look at recent changes to build the recent activity list and analyse what they mean which is not only a heavy operation but also dubious. Would be way better to just log actions like these in a list and read them from there. That is also the way minibiblio works and that would hopefully someday come to biblio.

PR #1666, opened 2024-07-24
Cleared curator role cookie on logout

### LINKED ISSUE #1390: Cookies not cleared on logout
# Bug description
On logout, all stateful information should be forgotten. The `biblio-backoffice` cookie is updated so the current user becomes "anonymous", but it does still contain information about the librarian mode (researcher or librarian). This produces problems if you log in as a researcher when the mode previously was set to librarian.

It may be best to just clear the cookie altogether, or at least to reprogram it to researcher mode on logout.

# Steps to Reproduce

1. Log in as a librarian. You enter in researcher mode.
2. Switch to librarian mode.
3. Log out again. Sidebar background should be blue but is gray.
5. Log in as researcher (you may have to clear keycloak cookies first to not be automatically logged on again)
6. You get the "Something went wrong" screen.

## Automatic testing scenario

I tried to write a failing test for this but couldn't make it work because of the specific hacky (interference) nature of Cypress of dealing with with cookies/sessions.

### Likelihood
Improbable

### Consequences
Marginal

# Expected behavior
Logging out should clear all state. Not just the user session.

# Screenshots
image

# Additional context
This may not currently occur in real use cases in TEST or DEV since Keycloak automatically logs you back on with the same account, but in prod you don't get a Keycloak screen in between and login is delegated to UGent login (CAS) automatically. So you could have a situation where two users work in the same browser and stumble upon this problem (after automatic or manual logout from CAS).

I recognise this is low priority but probably also very easy to fix.

verheyenkoen:
Also adapted curator login behaviour: curators are now always logged on in librarian mode but can still switch to researcher mode.

FIXES #1390

### DIFF
--- a/handlers/authenticating/handler.go
+++ b/handlers/authenticating/handler.go
@@ -6,6 +6,7 @@ import (
 
 	"slices"
 
+	"github.com/gorilla/sessions"
 	"github.com/ugent-library/biblio-backoffice/ctx"
 	"github.com/ugent-library/biblio-backoffice/vocabularies"
 	"github.com/ugent-library/bind"
@@ -48,7 +49,9 @@ func (h *AuthHandler) Callback(w http.ResponseWriter, r *http.Request) {
 	}
 
 	session.Values[ctx.UserIDKey] = user.ID
-	if _, ok := session.Values[ctx.UserRoleKey]; !ok {
+	if user.CanCurate() {
+		session.Values[ctx.UserRoleKey] = "curator"
+	} else {
 		session.Values[ctx.UserRoleKey] = "user"
 	}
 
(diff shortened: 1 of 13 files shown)


handlers/authenticating/handler.go
```diff
@@ -113,3 +112,11 @@ func UpdateRole(w http.ResponseWriter, r *http.Request) {
...
 	w.Header().Set("HX-Redirect", c.PathTo("dashboard").String())
 }
+
+func ClearSession(w http.ResponseWriter, r *http.Request, session *sessions.Session) error {
+	delete(session.Values, ctx.UserIDKey)
+	delete(session.Values, ctx.OriginalUserIDKey)
+	delete(session.Values, ctx.OriginalUserRoleKey)
+	delete(session.Values, ctx.UserRoleKey)
+	return session.Save(r, w)
+}
```
verheyenkoen, line 122: Moved this function over from handers/errors.go and made public, since they do the same thing now. I figured this file is a better place for that.

mietcls (commented):
Logging in and seeing the librarian dashboard immediately works as expected. I cannot seem to test the error page, but I presume that does not work locally.

nics (approved):

Issue #1613, opened 2024-06-17
Can no longer see project description in search


mietcls:
# Bug
Can no longer see project description in search
This was possible in previous dev deploy with old project configuration
<img width="1511" alt="Screenshot 2024-06-17 at 11 53 59" src="https://github.com/ugent-library/biblio-backoffice/assets/1572885/d648e4ad-e664-484a-8e33-7a3a7aab98ae">

The description is the abstract:
![screenshot_2024-06-27_at_10 37 54](https://github.com/ugent-library/biblio-backoffice/assets/1572885/9f3f3590-d680-4d0f-982f-168168a26c11)

nicolasfranck:
cf. https://github.com/ugent-library/biblio-backoffice/blob/dev/views/publication/suggest_projects.templ#L26
cf. https://github.com/ugent-library/biblio-backoffice/blob/dev/views/dataset/suggest_projects.templ#L26

The field description is taken into account.

On the master branch, I see that the `eu_acronym` was put there: https://github.com/ugent-library/biblio-backoffice/blob/main/backends/authority/project.go#L161

nicolasfranck:
@mietcls I forgot why this ticket was reopened?

mietcls:
Because the description (abstract) of the project is not shown. I'm not sure why.

For example: https://biblio.ugent.be/project/001D06803

In backoffice:
<img width="451" alt="Screenshot 2024-07-29 at 09 15 50" src="https://github.com/user-attachments/assets/097846b7-274b-496a-b579-e45db3e4cc9f">

Missing abstract:
<img width="1320" alt="Screenshot 2024-07-29 at 09 16 25" src="https://github.com/user-attachments/assets/37ec325b-ed92-4862-8916-161b65fbe30f">

nicolasfranck:
TODO:

* Description must become contents of abstract
* `eu_acronym` must be moved to list of identifiers (next to iweto_id)

PR #1671, opened 2024-07-29
fix issue #1613: put abstract in project description, and move eu acronym to list of identifiers

nicolasfranck:


nicolasfranck:
redo this again dev branch

PR #1674, opened 2024-07-30
fix issue #1613

nicolasfranck:


mietcls:
I can see the description now, great, please strip the description from the `<p></p>` tags like in the biblio front office.

PR #1681, opened 2024-07-30
Enhancement/plato enhancements

mietcls:
Incorporated some feedback gotten after the first Plato release

- Filters were not responsive enough for split screen
- Linking datasets to publications and vice versa looked different from projects, now aligned
- Added a dot to "last edit by"
- Improved copy and button positioning for the next Plato suggestions release
- Add suggested departments in overview

These tasks are part of #1623


views/dataset/add_publication_dialog.templ
```diff
@@ -30,6 +24,11 @@ templ AddPublicationDialog(c *ctx.Ctx, dataset *models.Dataset, publicationHits
 					hx-trigger="keyup changed delay:500ms"
 					hx-target="#publication-suggestions"
 				/>
+				<p class="text-muted">
+					Select one or more publications as already registered in Biblio. <a href="https://onderzoektips.ugent.be/en/tips/00002061/" target="_blank">View documentation <i class="if if--small if-external-link"></i></a>
```
verheyenkoen, line 28: Is the word "as" intended here?

views/publication/add_dataset.templ
```diff
@@ -30,6 +24,11 @@ templ AddDataset(c *ctx.Ctx, publication *models.Publication, hits *models.Datas
 					hx-trigger="keyup changed delay:500ms"
 					hx-target="#dataset-suggestions"
 				/>
+				<p class="text-muted">
+					Select one or more datasets as already registered in Biblio. <a href="https://onderzoektips.ugent.be/en/tips/00002061/" target="_blank">View documentation <i class="if if--small if-external-link"></i></a>
```
verheyenkoen, line 28: Same "as".

views/record_timestamp.go
```diff
@@ -26,5 +26,6 @@ func UpdatedBy(c *ctx.Ctx, updatedAt *time.Time, updatedBy *models.Person, lastU
 	if lastUpdatedBy != nil {
 		s += " Last edit by " + lastUpdatedBy.FullName
 	}
+	s += "."
```
verheyenkoen, line 29: Not sure this is right but can't test as this requires a system update to the record. Looks like if there is no last edit (`lastUpdatedBy == nil`), there will be two dots at the end.

verheyenkoen (changes requested):
Some tests are broken because of these changes. Should be easy to fix.

nics (approved):

Issue #1663, opened 2024-07-18
Add publication/dataset detail page sidebar


verheyenkoen:
# Enhancement
Add an icon sidebar on the right side in detail pages with 2 icons that open/toggle a drawer when clicked.

- Biblio messages (for librarians/curators this also contains reviewer notes and tags fields)  
  <img width="188" alt="image" src="https://github.com/user-attachments/assets/f8ba6688-f12b-41ba-a4c5-4b9d1f160634">

- Details
  - Author/editor name and date
  - Source (if applicable)
  - All identifiers with links to external sources  
  <img width="188" alt="image" src="https://github.com/user-attachments/assets/49a72169-6fc3-4a81-8e04-2c0b9bd82783">

[Prototype](https://biblio-backend-ugent-prototype.vercel.app/biblio/researcher/publications/detail/full-view.html)

Out of scope: Activity history

mietcls:
Added:

- librarian tags
- librarian note

Suggestion for save: buttons for now until we have a coherent approach to saving everywhere (otherwise we will have different patterns which will be confusing).

![Screenshot 2024-07-18 at 12 12 19](https://github.com/user-attachments/assets/6fb7f5f0-0553-479c-87d2-f57565fa182b)

mietcls:
Q: Can we remember whether the message tab was opened or closed by default?

verheyenkoen:
> Q: Can we remember whether the message tab was opened or closed by default?

<img src="https://github.com/user-attachments/assets/2a0ed887-8f3f-41eb-843e-a717afbbefb1" width="200">

verheyenkoen:
@nics I started this a couple of weeks ago but very quickly bumped into the fact that this is needed for both publications and datasets so the question is: is there something we can do to make sure we don't have to build the sidebar twice? We talked about deduping other views but never took any steps AFAIK. I guess it would be good to finally kick this off before we create any new duplicate views.

nics:
@verheyenkoen you can dedup them if it isn't too much work

verheyenkoen:
@nics I guess my question wasn't clear. How do we do that? I guess we need some kind of common structure for publications and datasets for that to be possible but I don't know how to achieve a base interface in Go that can be both. As I don't really know how this is done, I can't really estimate the effort for that.

nics:
[...]

PR #1682, opened 2024-07-31
Detail pages sidebar #1663

verheyenkoen:


mietcls:
Suggestion: on hold until as we need to focus on the LSP migration + implementation might be easier later.

PR #1686, opened 2024-08-06
Fixed linked publications/datasets visibility in detail pages #1372

### LINKED ISSUE #1372: Administrative support staff cannot see linked datasets/publications in a record if they are not assigned
# Bug description
Administrative support staff cannot see linked datasets/publications in a record if they are not assigned.
Needs to be fixed in main.

- [ ] When closing this: Miet needs to inform Evelien. Ping Miet

# Steps to Reproduce

1. Go to https://backoffice.biblio.ugent.be/publication/01HMBCF80GVRGEKWH815G4MM3M?redirect-url=%2Fpublication%3Fq%3D01HMBCF80GVRGEKWH815G4MM3M%26sort%3Ddate-updated-desc
2. See linked datasets
3. View as https://telefoonboek.prd.rad.ugent.be/nl/people/801002035845
4. You cannot see the datasets

View as librarian:
Screenshot 2024-02-07 at 17 29 52

View as [editor](https://telefoonboek.prd.rad.ugent.be/nl/people/801002035845):
Screenshot 2024-02-07 at 17 30 30

## Expected result

You can always see the linked datasets and publications inside a record that you are allowed to see, regardless of your role.

- rule 1: we always show the linked objects a user has the right to view, regardless of their current role
- rule 2: if there are linked objects that a user has no right to view, we show a version without links, you can only remove the object
- rule 3: we allow users to link objects if they have the right to edit the record they are currently editing and the right to view the object they are linking

I suggest not blocking the link to datasets/publications for now, so people can easily check whether it is the correct record that is linked.

verheyenkoen:
Implemented as described in #1372.

Implementation details (specifically for researchers that have now viewing/editing rights on the linked record):
* If this linked record has more than 3 authors resp. creators, there is no link behind the "x more author(s)/creator(s)", just plain text.
* If this linked record has no departments added, then there is also no "Add department" link in the card. Instead we display a plain text "No department(s)".
* For publications, if there is no main file, then the "Add document type: full text" link is replaced with "No document type" in plain text.

There are also an edge cases when the linked record is withdrawn and some fields are emptied again (which is not allowed while published):
* For publications, if all authors are removed we display "No author(s)" in plain text instead of the "Add UGent author" link.
* For datasets, if all creators are removed we display "No creator(s)" in plain text instead of the "Add UGent creator" link.
* For datasets, if the license is cleared we display "No license" instead of the "Add license" link.
* For datasets, if the access level is cleared we display "No document access level" instead of the "Add document access level" link.

These last cases are not Cypress tested as they are really specific and may never occur, but they were tested manually.

FIXES #1372


views/contributor/summary.templ
```diff
@@ -10,11 +10,13 @@ import (
...
-	URL              string
-	URLTarget        string
-	CurrentUserRoles string
+	Role                    string
+	Contributors            []*models.Contributor
+	URL                     string
+	URLTarget               string
+	CurrentUserRoles        string
+	CanViewMoreContributors bool
+	CanEditContributors     bool
```
verheyenkoen, line 19: Not the biggest fan of these 2 new flags, but don't know how we can do this in another manner when the contributor summary is not even aware of it's surrounding context (either publication or dataset).

verheyenkoen:
> There are also an edge cases when the linked record is withdrawn and some fields are emptied again (which is not allowed while published)

Just noted this is not as edgy as possible. If you link a dataset or publication to a draft publication resp. dataset, the reverse also shows a draft item, which can have these empty fields.

mietcls:
Only rule 3 isn't correct:
> we allow users to link objects if they have the right to edit the record they are currently editing and the right to view the object they are linking

However, now the users can see all of the datasets and publications, even if they don't have the right to see them.
In this case, I am logged in as researcher1, and I can see John Doe's publication – even if it is not linked yet:
Screenshot 2024-08-08 at 09 00 54

This behaves as expected:
rule 1: we always show the linked objects a user has the right to view, regardless of their current role
rule 2: if there are linked objects that a user has no right to view, we show a version without links, you can only remove the object

mietcls:
@verheyenkoen works as intended, rules need to be adapted not your PR, you can open it again!

mietcls (approved):

nics:
@mietcls I think this shows summary information even if a user has no permission to view the linked record (anymore)

mietcls:
@nics Yes indeed, but I saw it's the case as well on production. https://github.com/ugent-library/biblio-backoffice/issues/1372#issuecomment-2275656264

I'm trying to figure out why we decided this – probably because of delegation issues. We can retract this decision.

nics (approved):

PR #1687, opened 2024-08-07
Add clear button to filter fields #1673

### LINKED ISSUE #1673: Improve search field by adding clearing
# Enhancement

We are not using the full potential of the search input field, by using `type="text"` instead of `type="search"`

- [ ] Add type="search" to input field
  - Main overview / search view `views/search.templ`
  - Adding publications in dataset view `views/dataset/add_publication_dialog.templ`
  - Adding datasets in publication view `views/publication/add_dataset.templ`
  - Adding departments `views/publication/add_department.templ` and `views/dataset/add_department.templ`
- [ ] When using the clearing icon `x`, the search is refreshed

Screenshot 2024-07-30 at 09 51 01

verheyenkoen:
Added a little script to make this post the search form automatically when the clear button is clicked.

I've also changed some htmx-triggers in the htmx-driven forms where search fields were used. They acted on the `keyup` event which I've changed to the `input` event since that also covers pasting text ([docs](https://htmx.org/examples/active-search/#:~:text=Note%20that%20the%20keyup%20event%20could%20be%20used%20as%20well%2C%20but%20would%20not%20fire%20if%20the%20user%20pasted%20text%20with%20their%20mouse%20(or%20any%20other%20non%2Dkeyboard%20method).)).

FIXES #1673


nicolasfranck:
It also submits the form in this situation:
* the search input is empty
* you type something
* you clear it by clicking the clear icon

in this case there is no reason to submit anything as the input value hasn't changed

assets/js/ui/search.ts
```diff
@@ -0,0 +1,18 @@
+export default function initSearchFields(el: HTMLElement) {
+  el.querySelectorAll<HTMLInputElement>("input[type=search]").forEach(
```
nicolasfranck, line 2: It also submits the form in this situation:

* the search input is empty
* you type something
* you clear it by clicking the clear icon

in this case there is no reason to submit anything as the input value hasn't changed

Not sure if you can detect this during a search event

nicolasfranck: @mietcls  Is the type=search and clear button also needed in the add/update contributor form?

verheyenkoen: > @mietcls Is the type=search and clear button also needed in the add/update contributor form?

@nicolasfranck You specifically mean for the first name and last name filter fields here?

mietcls:
- Chrome: works
- Firefox: no `x` is shown
- Safari: `x` is shown but clearing action does not work

Not all browsers behave the same, so it's not a big deal as this was an enhancement:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#differences_between_search_and_text_types (open in chrome to test it on this website).

If we want this to behave the same in every browser, we will have to extend the native functionality.

mietcls (dismissed):

verheyenkoen:
@mietcls Implemented a fix that kinda works everywhere. It requires keeping track of the previous value and only submitting if that was not empty. That means the submit also immediately fires if you delete the last character in the field (with delete or backspace) which may not be what you want.

May be better to ask Emily to design a component for this with a button that we can simply put a click event on.

mietcls (changes requested):
I would leave out the clear on empty because it is non-standard behaviour – it might not be what you expected to happen and your focus leaves the search box as well.

For now standard browser behaviour is fine, even though not every browser behaves as you would want them to behave.

mietcls:
https://github.com/ugent-library/biblio-backoffice/pull/1687#pullrequestreview-2237504404 
Fixed with [b576b24](https://github.com/ugent-library/biblio-backoffice/pull/1687/commits/b576b249a5ada9743b4c68f5d7d12321a0bdfcce)

mietcls (approved):

assets/js/ui/search.ts
mietcls: Ah yes, that would be nice. I can create a separate issue so we don't have to wait to merge this one.

PR #1699, opened 2024-09-03
Strip HTML from project descriptions #1696

### LINKED ISSUE #1696: <p> and </p> tags showing in project description
# Bug Description

`<p>` and `</p>` tags showing in project description. We can't show this in production.

We should filter this out. We are aware there are an endless stream of other tags that could possibly filtered out, but we are going for the 80% problem, not the 100% problem for this fix.

Short term, we are solving `<p>` and `</p>` tags only.
Long term, we are asking GISMO to remove it (but they do not have focus for this now)

Screenshot 2024-08-30 at 09 12 59

verheyenkoen:
FIXES #1696

### DIFF
--- a/cypress/e2e/issues/issue-1696.cy.ts
+++ b/cypress/e2e/issues/issue-1696.cy.ts
@@ -1,0 +1,22 @@
+// https://github.com/ugent-library/biblio-backoffice/issues/1696
+
+describe("Issue #1696: <p> and </p> tags showing in project description", () => {
+  it("should strip HTML from project descriptions", () => {
+    cy.loginAsResearcher("researcher1");
+
+    cy.setUpPublication();
+    cy.visitPublication();
+
+    cy.contains(".card", "Project").contains(".btn", "Add project").click();
+
+    cy.ensureModal("Select projects").within(() => {
+      cy.setFieldByLabel("Search project", "001D07903");
+
+      cy.get("#project-suggestions .list-group-item")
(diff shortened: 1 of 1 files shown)


views/helpers.templ
```diff
@@ -1,48 +1,11 @@
...
-	}
-
-	after := text[pos:]
-	if len(after) > 0 {
-		b.WriteString(after)
-	}
-
-	return b.String()
-}
-
```
verheyenkoen, line 45: Moved this method to views/util along with StripHTML (that came from views/aria), see above. Two templ's still remain in this file (CSRFTag and Cat). Not sure if we should leave this like that, or give them their dedicated files.

views/aria/attributes.go
```diff
@@ -24,14 +23,3 @@ func Attributes(helpText string, helpFieldID string) templ.Attributes {
...
-func stripHTML(text string) string {
-	result, err := html2text.FromString(text, html2text.Options{})
-	if err != nil {
-		log.Printf("Error while stripping HTML from '%s': %s\n", text, err.Error())
-
-		return text
-	}
-
-	return result
-}
```
verheyenkoen, line 37: Moved this method (in an adapted version) to views/util to enable reuse. Could not add this to views/helpers.templ as that introduced circular references.

nics (approved):

nics (approved):

PR #1700, opened 2024-09-06
Fix contributor reorder concurrency conflict handling #1568

### LINKED ISSUE #1568: Missing conflict handling when editing / adding / removing contributors
# Bug description

When I add, remove or order a contributor from a publication or dataset which was being edited by someone else; I don't see a warning that a conflict error has happened. However, when I reload the page, I notice that my change weren't saved: I see the changes made by the other person who was editing instead.

# Steps to Reproduce

1. Open the detail page of a publication & go to the "people & affiliations" tab (A)
2. Add three new authors.
3. Duplicate the page in a separate browser tab (B)
4. Remove the authors in B.
5. Go back to A and reorder the 3 authors you're seeing.
6. Now add an author in A.
7.  Upon hitting the "save" button in the suggest box; notice how the table refreshes however your changes to the ordering are suddenly gone along with the authors who were just there a minute ago.

## Automatic testing scenario

Write a high-level way to test this with Cypress if applicable.

### Likelihood

Probable

### Consequences

Critical. If multiple people add, remove & reorder contributors at the same time, they'll see contributors inexplicably appear and disappear because there's no indication someone else is working on the same record at the same time.

# Expected behavior

All mutative actions should give you a conflict error dialog whenever someone else made a change.

verheyenkoen:
FIXES #1568


nics:
This doesn't really fix the concurrency issue. We should first check the snapshot id, and show a conflict error if someone else has modified the record. Checking the lengths is to validate the data posted and avoid out of bounds errors.

nics (approved):

Issue #1669, opened 2024-07-29
[Plato imports] Rejected / claimed suggestions


mietcls:
# Enhancement

Whenever I reject or claim a suggestion, a reduced version is shown of the suggestion.

[Reviewer suggestions view](https://booktower-ugent-ui.vercel.app/biblio/reviewer/reviewer-suggestions/reviewer-suggestions.html) (and [researcher suggestions view](https://booktower-ugent-ui.vercel.app/biblio/researcher/researcher-suggestions/researcher-suggestions.html) to be built):
![Screenshot 2024-07-30 at 10 27 35](https://github.com/user-attachments/assets/6affcb43-b9b5-4228-abf3-c4817cab84f3)

Researcher dashboard ([dashboard suggestions](https://booktower-ugent-ui.vercel.app/biblio/researcher/dashboard/dashboard.html) to be built):
![Screenshot 2024-07-30 at 10 28 03](https://github.com/user-attachments/assets/e142e17c-115d-4afe-8fc4-bd515f437b54)

## Claimed
![Screenshot 2024-07-30 at 10 28 29](https://github.com/user-attachments/assets/0844fb96-815c-4628-8581-e152cddd0548)
- [x] Status: `Claimed` + `source name` (in this case: Plato) + suggestion
- [x] Type (in this case: Dissertation)
- [x] Access level
- [x] Title
- [x] Summary of what happened to the record:
  - [x] `Claimed`
  - [x] Name of the person that claimed the record
  - [x] Action: `View` (opens record; only available for imported suggestions in next version)
  - [x] Date
  - [x] Countdown of how long the reminder will remain visible.

## Rejected
![Screenshot 2024-07-30 at 10 28 42](https://github.com/user-attachments/assets/9a1cd39b-7372-44aa-9507-03ea02ea7f1e)
- [x] Status: `Rejected` + `source name` (in this case: Plato) + suggestion
- [x] Type
- [x] Access level
- [x] Title
- [x] Summary of what happened to the record:
  - [x] `Rejected`
  - [x] Name of the person that rejected the record
  - [x] Action: `restore` (restores suggestion and refreshes list)
  - [x] Date
  - [x] Countdown of how long the reminder will remain visible.

## Sequence
- [x] Researcher dashboard: 
  - [x] top 4 newest import suggestions
  - [x] rejected and claimed suggestions are not displayed here, unless the ones just rejected from this view
[...]

PR #1708, opened 2024-09-12
Imported & rejected suggestions #1669

verheyenkoen:


cli/backends.go
```diff
@@ -342,6 +342,21 @@ func newRepo(conn *pgxpool.Pool, personService backends.PersonService, organizat
...
+					if err != nil {
+						logger.Warn("error loading status person in candidate record", "statuPersonID", *c.StatusPersonID, "id", c.ID, "error", err)
+						c.StatusPerson = backends.NewDummyPerson(*c.StatusPersonID)
+					} else {
+						c.StatusPerson = person
+					}
+				}
+				return nil
+			},
+		},
```
verheyenkoen, line 359: After calling this new CandidateRecordLoader, we always also call the unmarshall on the metadata field and then all PublicationLoaders on `rec.Publication`. I was looking to add those two routines as other CandidateRecordLoaders so this can all be done from one loop, but that didn't work since the PublicationLoader needs access to `Repo.config.PublicationLoaders` which isn't available from another loader I guess.

The unmarshall would be possible as a separate CandidateRecordLoader but not sure if that makes sense if doing the next step is still implemented in each repo method. Better to figure out a way to do it all as CandidateRecordLoaders.

db/query.sql
```diff
@@ -33,10 +34,15 @@ SELECT EXISTS(SELECT 1 FROM candidate_records WHERE status = 'new' AND (metadata
 
 
 -- name: GetCandidateRecord :one
-SELECT * FROM candidate_records WHERE status = 'new' AND id = $1 LIMIT 1;
+SELECT * FROM candidate_records WHERE id = $1 LIMIT 1;
```
verheyenkoen, line 37: I had to remove the `status = 'new'` condition here otherwise the `SetCandidateRecord` router middleware for restoring rejected suggestions would fail with a 404. I could make this `... WHERE status IN ('new', 'rejected') AND ...` but not sure if that doesn't make it more confusing by filtering specific statuses in a method with a generic name.

assets/ugent/scss/components/_c-button-toolbar.scss
mietcls, line 1: Nice, can you adapt this in the prototype all well? Thanks!

sqlc.yaml
```diff
@@ -8,3 +8,4 @@ sql:
         package: "db"
         out: "db"
         sql_package: "pgx/v5"
+        emit_pointers_for_null_types: true
```
verheyenkoen, line 11: This was necessary to get the mapping working. Nullable text fields are now generated as `*string` in the repo. Otherwise it uses specific Pg-types but couldn't get that to work properly.

views/candidaterecord/summary.templ
```diff
@@ -5,42 +5,86 @@ import (
...
-	Badge     templ.Component
+	Thumbnail   string
+	Badge       templ.Component
+	Info        templ.Component
+	ShowDetails bool
+}
+
+func daysUntilDisappearanceDate(date time.Time) string {
+	disappearanceDate := util.DateOnly(date).AddDate(0, 0, 90)
+	today := util.DateOnly(time.Now())
```
nics, line 22: Can't you simply `time.Now().Round(0).Date()`? (haven't tested)

verheyenkoen: That doesn't work according to the [docs](https://pkg.go.dev/time#Time.Round) and [my test case](https://goplay.tools/snippet/LVtaF9PHQvE).

You could round to a date using `.Round(time.Hour * 24)` I think but that would round up to the next day after 12:00 which is also not what we want.

assets/ugent/scss/components/_c-button-toolbar.scss
verheyenkoen: Btw, already did this but forgot to mention here: https://github.com/ugent-library/biblio-ugent-prototype/pull/22

sqlc.yaml
verheyenkoen: Fixed in c3142283c8ad7b63b1a5f80815999ed9e4ea373d

PR #1731, opened 2024-10-22
Restrict restore of rejected suggestions by biblio team members #1728

### LINKED ISSUE #1728: [Plato Imports] Make removed suggestion more nuanced: duplicate
# Enhancement
[Updated content]

1. Make removed suggestion for Plato more clear: rejected as a duplicate.
Other rejection options will be needed for other sources, here we know there is little probability we are suggesting it to the wrong person.

Screenshot 2024-11-15 at 09 48 49
Screenshot 2024-11-15 at 09 48 40

2. Hide "View" after import for researchers for now
When you import a record, and then delete the publication, the view button does not work. This causes confusion. We will fix that later for all researchers, but now keep the button for reviewer workflows.

verheyenkoen:
FIXES #1728

- **Makes sure "rejecting" is now communicated as "rejected as a duplicate"**.
We will add other rejection reasons in the future for other sources, for Plato there will be little other reasons.
Restoring remains possible for all parties.
- **Make sure "View" is hidden for researchers for now.**
Researchers were able to delete a publication after import, which breaks the link. This has caused confusion with our proxy testers.
This is now hidden for researchers and proxies, but we kept it for workflow purposes for reviewers.

### DIFF
--- a/handlers/candidaterecords/candidate_records.go
+++ b/handlers/candidaterecords/candidate_records.go
@@ -115,6 +115,16 @@ func RestoreRejectedCandidateRecord(w http.ResponseWriter, r *http.Request) {
 	c := ctx.Get(r)
 	rec := ctx.GetCandidateRecord(r)
 
+	if rec.Status != "rejected" {
+		c.HandleError(w, r, httperror.BadRequest)
+		return
+	}
+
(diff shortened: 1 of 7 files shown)


mietcls (approved):
I only improved a bit of the copy, works as expected, thanks.

mietcls (approved):

mietcls:
Review by @mietcls needed – remove for everyone when it is removed by a curator.

PR #1735, opened 2024-11-06
Fix expired embargoes on candidate records #1726

### LINKED ISSUE #1726: [Plato imports] embargo date incorrect
# Bug description
Plato imports: embargo date does not switch

# Steps to Reproduce

1. Go to suggestions: https://backoffice.biblio.ugent.be/candidate-records
2. See imported plato suggestion, with embargo date still active:
SCR-20241014-ouje
- e.g. Towards the Next Generation of Geometric Phase Photonic Components Using Chiral Liquid Crystal
- Open access from 2024-10-01
- Claimed on On 2024-10-10
3. See import itself Screenshot 2024-10-14 at 16 51 13 (changed manually by Biblio team member)

## Automatic testing scenario
Possible @verheyenkoen ?

## Expected behavior
The "switch" for the embargo happens in the suggestion itself, just like for the ones we create ourselves in Biblio.
The Biblio team does not need to see that information in the suggestion if the embargo is already passed.

### Likelihood
How often does this happen?
3: Occasional

### Consequences
2: Critical

verheyenkoen:
FIXES #1726


repositories/candidate_records.go
```diff
@@ -398,3 +397,56 @@ func (r *Repo) loadCandidateRecord(result *models.CandidateRecord) error {
 
 	return nil
 }
+
+func (r *Repo) UpdateCandidateRecordEmbargoes() (int, error) {
+	var n int
+
+	psql := sq.StatementBuilder.PlaceholderFormat(sq.Dollar)
+	query := psql.Select("*", "-1 AS total").
```
verheyenkoen, line 405: Don't like this `-1 AS total` but it's kinda necessary to reuse the `queryRows` function below. Tried many ways but couldn't find a way to make this logic more generic.

nics: pretty it ain't, but I get why...

nics (approved):

mietcls (approved):

Issue #1739, opened 2024-11-15
[Plato imports] Fix confusing copy in suggestions preview


mietcls:
# Enhancement
Fix confusing copy in suggestions preview
- 4 questions: "-" is unclear
- "Dut" is unclear
- Also unclear to some people that some fields need to be reviewed and completed.

PR #1740, opened 2024-11-15
Enhancement/suggestions copy fixes

mietcls:
Fixes https://github.com/ugent-library/biblio-backoffice/issues/1739

- Would be better to use "localize.LanguageName" for the abstract language, needs to happen later.


views/publication/details.templ
```diff
@@ -226,18 +226,22 @@ templ DetailsBody(c *ctx.Ctx, p *models.Publication) {
...
 					})
 					@display.Field(display.FieldArgs{
 						Label: c.Loc.Get("builder.has_publications_planned"),
 						Value: c.Loc.Get("confirmations." + p.HasPublicationsPlanned),
+						Required: true,
 					})
 					@display.Field(display.FieldArgs{
 						Label: c.Loc.Get("builder.has_published_material"),
 						Value: c.Loc.Get("confirmations." + p.HasPublishedMaterial),
+						Required: true,
```
verheyenkoen, line 244: I assume it was decided to just add the "Required" label but don't enforce it? Otherwise, I'm missing validation logic for this.

verheyenkoen (approved):

views/publication/details.templ
mietcls: Yes, we're not enforcing it, but mostly getting people's attention to at least say "I don't know" if they don't.
I am working on a better version of this with Marthe, but in the meantime to support the current reviewers workflow we added this.

verheyenkoen (approved):

verheyenkoen:
Tested. Looks good.

nics (approved):

Issue #1775, opened 2026-06-04
batch operations to set_publisher and remove_publisher


mietcls:
straight forward, similar to:
```
			"set_publisher":                    mutate.SetPublisher,
			"remove_publisher":              mutate.RemovePublisher,
```

nicolasfranck:
@mietcls in production. Can you update the instructions at https://booktower.gitbook.io/product-docs/producten-en-diensten/biblio-academische-bibliografie-en-repository/handleidingen/batch-operaties#maak-het-jezelf-gemakkelijk ?

```
<id>,set_publisher,<publisher>
<id>,remove_publisher
```

mietcls:
Done, is this correct?

<img width="546" height="319" alt="Image" src="https://github.com/user-attachments/assets/bf806f6a-6417-4d10-b5d6-c83058a289a5" />

https://booktower.gitbook.io/product-docs/producten-en-diensten/biblio-academische-bibliografie-en-repository/handleidingen/batch-operaties#maak-het-jezelf-gemakkelijk

Thanks!

nicolasfranck:
For `remove_publisher` it is `<id>,remove_publisher` (there is no third argument, as it merely sets one argument to an empty string).

PR #1776, opened 2026-06-05
implement mutate.SetPublisher and mutate.RemovePublisher

nicolasfranck:
Fixes #1775 

* adds mutation `set_publisher`
* adds mutation `remove_publisher`
* adds tests for both functions. I used the golang's builtin testing.T suit. I see that in `client/` another one is used. Harmless?


nicolasfranck:
@nics merged into dev on request of @mietcls to test on bibliodev

